PR #50, opened 2026-05-19
feat(github): connect commit SHAs in traces to GitHub via a GitHub App

JeremyFunk:
## Summary

Adds a GitHub integration so the `deployment.commit_sha` resource attribute on incoming spans resolves to author, message, avatar, and a GitHub deep link — across the services table, trace headers, span detail panels, attribute tables, the filter sidebar, and (via `AttributeRow` auto-detection) anywhere else attributes are rendered.

Supports **multi-installation across multiple GitHub orgs and personal accounts**, syncs **all branches** (default branch backfilled on connect, others picked up via `push` webhooks), and works out of the box on both Cloud Maple and self-hosted deploys.

### What we added

- **5 new D1 tables** (`packages/db/drizzle/0014_github_integration.sql`): `github_installations`, `github_repositories`, `github_commits`, `github_releases` (groundwork only — schema in place, no sync logic yet), `github_unresolved_shas` (DB-side tombstone for SHAs we tried to resolve and failed).
- **4 backend services**:
  - `GithubAppJwtService` — RS256 JWT minting + installation token cache + webhook HMAC signature verification (constant-time via Web Crypto).
  - `GithubInstallationClient` — REST client with `Link: rel="next"` pagination, 404/422/409 normalization, per-installation rate-limit handling.
  - `GithubAppService` — install/disconnect/list orchestration; **hard-delete cascade** on disconnect (installation + repos + commits + releases + tombstones) so re-connecting starts from a clean slate.
  - `GithubSyncService` — the 4 job handlers: `runBackfill` (paginated, queue-resumable), `runWebhookPush` (commits from push payload or `compare` fallback), `runResolveUnknownSha` (best-effort SHA lookup across connected repos, tombstone after 3 failures), `runReconcile` (refresh installation metadata + repo list).
- **2 new HTTP API groups**: `integrations.github.*` (install flow, list installations/repos, set sync, backfill, disconnect) and `commits.*` (batched SHA lookup + manual resync).
- **UI components**: `CommitChip` with hover-card (author avatar, message, GitHub link), `CommitLookupProvider` for route-level SHA batching, `GitHubIntegrationCard` for settings, plus 3 new icons.
- **`AttributeRow` auto-detection** — 7 commit-SHA attribute keys (`deployment.commit_sha`, `git.commit.sha`, etc.) automatically render as `CommitChip` whenever an attribute table is shown. This means span detail, log detail, future error/alert panels — anywhere attributes appear — get the chip for free.

### Why Cloudflare Queues

The four job types are all async work the user's request shouldn't block on:

- **Initial backfill** is ~10 paginated GitHub API calls (90 days of commits on the default branch). Inline would block the install callback for several seconds.
- **Webhook push handling** must ack GitHub within 10s; offloading to a queue keeps the webhook endpoint fast and lets us retry on transient failures.
- **Unknown-SHA resolution** fires opportunistically when a user hovers an unresolved chip. Queue means we don't block the hover render.
- **6h reconcile cron** scans all active installations to catch missed webhooks.

We considered:
- **`ctx.waitUntil`** — works for happy-path inline calls but no durability, no retries, and 30s CPU ceiling per request can truncate large backfills.
- **Cloudflare Workflows** — durable multi-step orchestration, but overkill for our linear page-by-page loops and adds a new mental model.

Queues hit the sweet spot: durable retries with exponential backoff, no per-job CPU ceiling, and lets `ctx.waitUntil` callsites stay simple. The queue's batch handler is the same Worker as the HTTP API (same `MainLive` layer, no version-skew between producer and consumer).

We **intentionally do NOT configure a dead-letter queue**. All four job types have application-level recovery paths (the 6h reconcile cron re-fetches missed commits, the tombstone table tracks failed SHA lookups, the UI Backfill button can re-trigger any repo). A queue-level DLQ would only be an inspection aid, at the cost of forcing operators to provision a second queue. After 5 retries, messages are dropped and recovery falls to the application layer.

### Infrastructure as code

The queue + cron + env-var bindings are declared in `apps/api/alchemy.run.ts`. **Running `alchemy deploy` provisions everything idempotently — no manual `wrangler queues create` step required**. Local dev (`bun dev`) uses miniflare's built-in queue simulator via `wrangler.jsonc`.

### For Cloud Maple

None at the application level — `alchemy deploy` handles infrastructure. Operations only needs to set 4 secrets in the deploy environment:

- `GITHUB_APP_ID`
- `GITHUB_APP_SLUG`
- `GITHUB_APP_PRIVATE_KEY` (multi-line PEM)
- `GITHUB_APP_WEBHOOK_SECRET`

Plus register the Maple GitHub App once on github.com with webhook URL `https://<app-domain>/api/webhooks/github` and setup URL `https://<app-domain>/api/integrations/github/callback`.

### For self-hosted deployments

Documented in `docs/github-integration.md`:
1. Operator registers their own GitHub App (or uses GitHub Enterprise Server).
2. Sets the same 4 env vars.
3. Runs `alchemy deploy`.

That's it. The queue, cron trigger, and all bindings come up automatically.

### Webhook event subscriptions on the GitHub App

Only `Push`, `Release` (groundwork), and `Pull request` (groundwork) need to be checked. `installation` and `installation_repositories` are delivered automatically to every GitHub App.

## Out of scope for v1 (groundwork ready, not implemented)

- **GitHub Releases sync** — `github_releases` table + `release` webhook subscription in place; sync logic + UI to follow.
- **PR number enrichment** — `github_commits.pr_number` column opportunistically populated; no UI surface yet.
- **Recharts release-marker labels** — service-detail chart still shows static SHA text on reference lines (the other 7 SHA surfaces all use `CommitChip`).

## Test plan

- [ ] \`bun run typecheck\` passes across all packages
- [ ] \`bun run test\` — 87 new tests added in 8 files:
  - \`packages/domain/src/queues/github-jobs.test.ts\` (8)
  - \`apps/api/src/services/GithubAppJwtService.test.ts\` (13)
  - \`apps/api/src/services/GithubSyncQueue.test.ts\` (9)
  - \`apps/api/src/services/GithubAppService.test.ts\` (16)
  - \`apps/api/src/services/GithubSyncService.test.ts\` (14)
  - \`apps/api/src/services/GithubInstallationClient.test.ts\` (10)
  - \`apps/web/src/components/commits/commit-chip.test.tsx\` (5)
  - \`apps/web/src/components/attributes/attributes-table.test.tsx\` (12)
- [ ] Local end-to-end (with a personal-account dev GitHub App + cloudflared tunnel):
  - Install flow completes, settings card shows installation + repo list
  - Push a commit → row appears in \`github_commits\` within seconds
  - Hover a commit chip in \`/services\` → tooltip shows author + message + GitHub link
  - Hover an unresolved SHA → tombstone tracked in D1
  - Disconnect → all rows for that installation cleaned up
- [ ] Migration \`0014_github_integration.sql\` applies cleanly on a fresh D1


apps/api/src/queues/cron-scheduler.ts
```diff
@@ -0,0 +1,32 @@
...
+// job for every active installation — catches webhooks GitHub failed to
+// deliver, refreshes repo lists, and re-syncs installation metadata.
+export const runScheduledReconcile = Effect.gen(function* () {
+	const database = yield* Database
+	const queue = yield* GithubSyncQueue
+	const installations = (yield* database
+		.execute((db) =>
+			db.select().from(githubInstallations).where(isNull(githubInstallations.suspendedAt)),
+		)
+		.pipe(Effect.orDie)) as ReadonlyArray<GithubInstallationRow>
```
Makisuo: fyi orDie is kinda forbidden its like unwrap in rust kinda

apps/api/src/routes/commits.http.ts
```diff
@@ -0,0 +1,159 @@
...
+
+const SHA_REGEX = /^[0-9a-f]{7,40}$/i
+
+const toPersistenceError = (cause: unknown) =>
+	new IntegrationsPersistenceError({
+		message: cause instanceof Error ? cause.message : "Commits lookup database error",
+	})
+
+const parseBranches = (json: string): ReadonlyArray<string> => {
+	try {
```
Makisuo: never use try catch in effect land use Effect Schema for stuff like this its very powerful

apps/api/src/routes/commits.http.ts
```diff
@@ -0,0 +1,159 @@
...
+} from "@maple/domain/http"
+import { and, eq, inArray } from "drizzle-orm"
+import { Effect } from "effect"
+import { HttpApiBuilder } from "effect/unstable/httpapi"
+import { Database, type DatabaseClient } from "../services/DatabaseLive"
+import { GithubSyncQueue } from "../services/GithubSyncQueue"
+
+const SHA_REGEX = /^[0-9a-f]{7,40}$/i
+
+const toPersistenceError = (cause: unknown) =>
```
Makisuo: this is already kind of a hack since it maps a cause should rather create some sort of client repo

apps/api/src/routes/commits.http.ts
```diff
@@ -0,0 +1,159 @@
...
+const chunk = <T>(xs: ReadonlyArray<T>, size: number): ReadonlyArray<ReadonlyArray<T>> => {
+	if (xs.length <= size) return [xs]
+	const out: T[][] = []
+	for (let i = 0; i < xs.length; i += size) {
+		out.push(xs.slice(i, i + size))
+	}
+	return out
+}
+
+export const HttpCommitsLive = HttpApiBuilder.group(MapleApi, "commits", (handlers) =>
```
Makisuo, line 15: probbaly commits is not the best Route name here would rather be somehtign liek github or something or idk, just somethign github speicifc

apps/api/src/routes/commits.http.ts
```diff
@@ -0,0 +1,159 @@
...
+	Effect.gen(function* () {
+		const database = yield* Database
+		const queue = yield* GithubSyncQueue
+
+		const dbExecute = <T>(fn: (db: DatabaseClient) => Promise<T>) =>
+			database.execute(fn).pipe(Effect.mapError(toPersistenceError))
```
Makisuo: this is kind of a hack but fine, also use it in some palces

apps/api/src/routes/commits.http.ts
```diff
@@ -0,0 +1,159 @@
...
+
+		return handlers
+			.handle("commitsLookupBySha", ({ payload }) =>
+				Effect.gen(function* () {
+					const tenant = yield* CurrentTenant.Context
+					const validShas = Array.from(
```
Makisuo: could maybe use a Effect Hashmap here, probably easier but could be wrong

apps/api/src/routes/commits.http.ts
```diff
@@ -0,0 +1,159 @@
...
+					)
+					if (validShas.length === 0) {
+						return new CommitsLookupResponse({
+							entries: payload.shas.map(
+								(sha) => new CommitsLookupEntry({ sha, commit: null }),
+							),
+						})
+					}
+					const commits: GithubCommitRow[] = []
+					for (const batch of chunk(validShas, 50)) {
```
Makisuo: never use for loops here in effect use Effect.forEach or similar since that way you can tune conccurancy

apps/api/src/routes/commits.http.ts
```diff
@@ -0,0 +1,159 @@
...
+							db
+								.select()
+								.from(githubRepositories)
+								.where(
+									and(
+										eq(githubRepositories.orgId, tenant.orgId),
+										inArray(githubRepositories.id, repoIds),
+									),
+								),
+						)) as ReadonlyArray<GithubRepositoryRow>
```
Makisuo: as types are kind of a hack imo satisfies is fine though

apps/api/src/routes/github-webhook.http.ts
```diff
@@ -0,0 +1,231 @@
...
+					return HttpServerResponse.text("unknown installation", { status: 200 })
+				}
+
+				const action = (body as { action?: string }).action
+
+				// Webhook delivery is bounded and time-sensitive (10s GitHub timeout).
+				// We run reconciliation inline rather than enqueueing because the
+				// payload is small, the user expects immediate UI updates after
+				// connect/push, and the 6h cron sweep catches anything we miss.
+				switch (event) {
```
Makisuo: Effect.match instead

Makisuo (changes requested):

PR #75, opened 2026-06-06
fix(docker): build workspace deps before web bundle + bump bun to 1.3.11

dbrimley:
## Summary

- `@maple-dev/browser` and `@maple-dev/effect-sdk` both export from `dist/` which is gitignored. Without an explicit build step, Rolldown can't resolve their imports and `docker compose up --build` fails for the web service with:
  ```
  Error: [vite]: Rolldown failed to resolve import "@maple-dev/browser" from "main.tsx"
  ```
- Bumps the bun base image from `1.3.8` to `1.3.11` across `api` and `web` Dockerfiles
- Removes the `otel-collector` `depends_on` from the `api` service (causes a circular startup problem)
- Adds `docker:build` and `docker:up` convenience scripts to root `package.json`

## Test plan

- [ ] `docker compose build` completes without error
- [ ] `docker compose up` starts all services
- [ ] `curl http://localhost:3471` returns `200 OK` from the web service

### DIFF
--- a/apps/api/Dockerfile
+++ b/apps/api/Dockerfile
@@ -1,5 +1,5 @@
 # Build context: monorepo root (.)
-FROM oven/bun:1.3.8 AS base
+FROM oven/bun:1.3.11 AS base
 WORKDIR /app
 
 FROM base AS pruner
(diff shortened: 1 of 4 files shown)


Makisuo:
Thanks!

PR #77, opened 2026-06-08
WIS-236: fix local receiver Java agent OTLP protobuf decode

bitzhjr:
## Summary
- upgrade `apps/cli` to `protobufjs@8.6.1` so the local receiver can decode OpenTelemetry Java agent OTLP protobuf payloads
- add a regression test with a wire-captured Java agent `POST /v1/traces` payload that previously threw `RangeError: index out of range`
- assert the decoded payload still encodes into Maple local `traces` rows with the expected service/scope/span fields

## Testing
- `bun run test -- src/server/otlp/encode.test.ts`
- `bun run test`

## Notes
- I attempted a local-mode end-to-end run from source as well. This checkout first needed `@maple-dev/effect-sdk/server` built, and after providing `libchdb` the server bootstrap still failed on an existing chDB/schema issue (`upperUTF8` missing during startup). The regression fixture in this PR is the real Java agent payload that was returning `400 decode traces` before the dependency upgrade.

### DIFF
--- a/apps/cli/package.json
+++ b/apps/cli/package.json
@@ -16,7 +16,7 @@
 		"@maple/domain": "workspace:*",
 		"@maple/query-engine": "workspace:*",
 		"effect": "catalog:effect",
-		"protobufjs": "^7.4.0"
+		"protobufjs": "^8.6.1"
 	},
 	"devDependencies": {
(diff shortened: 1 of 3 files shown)


Makisuo:
Looks good, thanks for the fix!

PR #93, opened 2026-06-14
Fix ClickHouse-ready ingest routing cache

Makisuo:
## Summary
- Split ingest-key/Cloudflare connector identity caching from per-org ClickHouse routing readiness caching.
- Add a short `INGEST_ORG_ROUTING_CACHE_TTL_SECS` default (1s) so setup changes stop being pinned behind the 60s auth cache.
- Keep non-ready ClickHouse orgs on the shared Tinybird forward path.
- Add a request-path regression proving a ready org in `INGEST_WRITE_MODE=forward` writes to ClickHouse without forwarding again.

## Tests
- `cargo test` in `apps/ingest`
- `cargo test forward_mode_switches_ready_org_to_clickhouse_without_forwarding_again` in `apps/ingest`
- `bun run clickhouse:schema:check`

### DIFF
--- a/apps/ingest/src/main.rs
+++ b/apps/ingest/src/main.rs
@@ -108,6 +108,8 @@ struct AppConfig {
     autumn_flush_interval_secs: u64,
     autumn_enforce_limits: bool,
     autumn_check_cache_ttl_secs: u64,
+    ingest_key_cache_ttl_secs: u64,
+    org_routing_cache_ttl_secs: u64,
 }
 
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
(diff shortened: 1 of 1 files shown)


Makisuo:
Addressed the review feedback in 5f6def4be. I removed the dead self-managed forward endpoint path entirely so Tinybird fallback always uses the shared managed collector, and ClickHouse-ready orgs stay on the direct ClickHouse native path. I also changed routing refresh failures to serve last-known routing when available, or managed Tinybird defaults when there is no known state, so a D1 routing blip does not 503 warm auth traffic or silently erase a known-ready ClickHouse route.\n\nValidation: `cargo test` in `apps/ingest`, `bun run clickhouse:schema:check`, and the PR GitHub checks are green.

PR #94, opened 2026-06-16
fix: render conditionally some components using autumn sdk

abusarah-tech:
This PR should fix the issue when `VITE_MAPLE_AUTH_MODE` is set to `self_hosted`. Some components require the use of Autumn SDK that is why it causes an error.


Makisuo:
Thanks looks really good!

PR #98, opened 2026-06-19
fix(query-engine): make raw-traces filters coalesce HTTP semconv like trace_list_mv

JeremyFunk:
## Problem

Trace **quick-filter (facet) counts** come from `trace_list_mv`, whose materialized-view SQL coalesces both OTel HTTP semconv spellings when it pre-extracts columns:

- `http.method` ↔ `http.request.method`
- `http.status_code` ↔ `http.response.status_code`

…and rewrites HTTP server span names to their display form (`"http.server GET"` + `http.route`/`url.path` → `"GET /api/users"`).

But **applying** a facet filters the raw `traces` table, which previously matched only the single literal attribute key / raw span name. Local data (chDB / a fresh Tinybird) emits only the **new** semconv keys and stores **raw** span names — so a facet would show a non-zero count, yet selecting it matched **zero rows**. This is the documented "facets vs list-filter invariant" break, most visibly the **"Root Span"** quick filter returning nothing locally.

## Fix

Mirror the MV's normalization on the raw-traces query path, all within `packages/query-engine`:

- **`traces-shared.ts`** — add `HTTP_SEMCONV_ALIASES` plus `coalescedMapGet` (`if(map[k0] != '', map[k0], map[k1])`) and `anyMapContains`; add the shared `httpDisplaySpanName(spanName, route, urlPath)` rewrite. `buildAttrFilterCondition` now coalesces both spellings for `SpanAttributes`.
- **`query-helpers.ts`** — `tracesBaseWhereConditions`: the `spanName` include branch and the `excludedSpanNames` exclude branch are now display-name aware, matching the raw **OR** rewritten span name.
- **`errors.ts`** — `spanHierarchyQuery` uses the shared `httpDisplaySpanName` instead of its inline copy.

## Scope & verification

Self-contained to `packages/query-engine`. The `trace_list_mv` side already coalesces both spellings on `main`; no datasource/MV/route companion change required.

`ch.test.ts` updated and passing (103/103) with assertions for the coalesced `if(... != '', ...)` SQL and the display-name-aware `NOT IN`.

> Extracted out of the `feat/github-integration` branch and rebased onto current `main` (which had refactored the CH DSL into `@maple-dev/clickhouse-builder`); one cosmetic conflict in `spanHierarchyQuery` was resolved in favor of `main`'s formatting.

### DIFF
--- a/packages/query-engine/src/ch/ch.test.ts
+++ b/packages/query-engine/src/ch/ch.test.ts
@@ -458,7 +458,10 @@ describe("tracesTimeseriesQuery", () => {
 			attributeFilters: [{ key: "http.status_code", value: "200", mode: "equals" }],
 		})
 		const { sql } = compileCH(q, baseParams)
-		expect(sql).toContain("SpanAttributes['http.status_code'] = '200'")
+		// Coalesces the old + new OTel semconv spellings (mirrors trace_list_mv).
+		expect(sql).toContain(
+			"if(SpanAttributes['http.status_code'] != '', SpanAttributes['http.status_code'], SpanAttributes['http.response.status_code']) = '200'",
+		)
 	})
 
 	it("filters by attribute filters (exists)", () => {
@@ -488,7 +491,9 @@ describe("tracesTimeseriesQuery", () => {
 			attributeFilters: [{ key: "http.status_code", value: "400", mode: "gt" }],
 		})
 		const { sql } = compileCH(q, baseParams)
-		expect(sql).toContain("toFloat64OrZero(SpanAttributes['http.status_code']) > 400")
+		expect(sql).toContain(
+			"toFloat64OrZero(if(SpanAttributes['http.status_code'] != '', SpanAttributes['http.status_code'], SpanAttributes['http.response.status_code'])) > 400",
+		)
 	})
 
 	it("filters by resource attribute filters", () => {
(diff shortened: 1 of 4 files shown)


JeremyFunk:
Addressed the drift-guard feedback in 4639ff7b — added the guard **in this PR** (not a follow-up).

**`packages/query-engine/src/traces-shared.test.ts`** compiles `httpDisplaySpanName` and `buildAttrFilterCondition` and asserts their SQL fragments appear in *both* downstream encodings, imported from `@maple/domain`:
- the `trace_list_mv_mv` materialized view (`materializations.ts`)
- ClickHouse migration `0004`'s backfill `select`

Comparison is whitespace/paren-insensitive (the CH DSL emits one redundant outer paren around the `if(...)` condition that the hand-written MV SQL omits — purely cosmetic; exact compiled SQL incl. parens is still pinned by `ch.test.ts`). A non-vacuous-pass sanity assertion covers the fragment extraction, and I verified the guard **fails on a simulated drift** of the `'http.server '` literal (3 parity assertions go red), then passes again on revert.

Also added cross-reference comments in `materializations.ts` and migration `0004` pointing at the TS mirror, so an editor of either SQL site knows to update all three.

`bun test src/traces-shared.test.ts src/ch/ch.test.ts` → 110 pass.

PR #105, opened 2026-06-20
feat(charts): drag-to-zoom, fullscreen expand, timezone-aware time controls

JeremyFunk:
Time-range UX for the dashboard, service-detail, and home-overview charts. (Arrow-key keyboard navigation is split into a follow-up PR stacked on this one.)

## What's included
- **Drag-to-zoom** — drag across any time-series chart to narrow the window to that selection. Click vs drag is decided by pointer travel (≥6px), so a plain click never zooms. Shared `useChartDragZoom` hook drives all chart types; read-only history previews don't mutate the live range.
- **Expand to fullscreen** — a maximize button (revealed on card hover) opens any chart in a near-fullscreen centered modal (`ChartExpandModal`). Also fixes the pre-existing widget-header hover-reveal (a `group/card` class was never declared, so actions only appeared after a click) and the action-button alignment.
- **Timezone selector** — the time-range picker footer is now a searchable popover: search by zone name **or** UTC offset, "System" and "UTC" pinned first, sorted, with aligned offset columns.
- **Timezone-correct chart times** — x-axis ticks, tooltips, and deploy-marker times render in the user's **selected** timezone (threaded through `formatBucketLabel`), not the browser's. Incoming buckets are normalized to UTC upstream, so re-projection is correct for any source offset.
- **Custom-range fix** — selecting a custom range no longer jumps by the UTC offset (timezone-aware round-trip); a reversed start/end is ordered ascending.

## Notes
- Base PR of a stack; the **keyboard navigation** PR (arrow-key pan/zoom + `F`-to-expand) targets this branch.
- Verified end-to-end in-browser across all three surfaces. `bun typecheck` at the branch baseline (web 23 pre-existing, ui 0); lint/format clean.


Makisuo:
@pullfrog resolve conflict

PR #106, opened 2026-06-20
feat(charts): keyboard navigation for time-series charts

JeremyFunk:
Keyboard control over the selected time window. **Stacked on #105** (the expand modal that `F` opens, plus the shared `isOverlayOpen` guard, live there) — review/merge #105 first; this PR's diff is keyboard-only.

## Keys (dashboard, service-detail, home)
| Key | Action |
|-----|--------|
| ← / → | Pan the window into the past / future (→ clamps at *now*) |
| ↑ / ↓ | Zoom in / out around the window center (min-width + ~2y lookback clamps) |
| Shift | Much larger step |
| Ctrl / Meta | Much finer step |
| F | Open the hovered chart in the fullscreen expand modal |

## Implementation
- `useTimeRangeKeyboardControls` — a `window` capture-phase keydown listener that operates on the resolved absolute window and writes an absolute range back. It `preventDefault`s only the keys it handles, and bails when an editable element is focused or an overlay (dialog / menu / listbox via `isOverlayOpen`) owns the keyboard — so it never hijacks menu/select/popover navigation.
- `F` is handled per widget-shell (only the hovered chart responds).

## Notes
- Verified end-to-end: pan, zoom, Shift/Ctrl magnitudes, right-clamp-to-now, and the overlay guard (arrows + F don't fire while the picker is open).
- `bun typecheck` at baseline (web 23 pre-existing, ui 0); lint/format clean.


apps/web/src/components/dashboard-builder/widgets/widget-shell.tsx
```diff
@@ -81,8 +82,37 @@ export function WidgetShell({
...
+	// leave) so only that one widget responds — no shared hover registry needed.
+	const [hovered, setHovered] = useState(false)
+	const expandRef = useRef<() => void>(() => undefined)
+	expandRef.current = () => {
+		if (canExpand) setExpanded(true)
+	}
+
+	useEffect(() => {
+		if (!hovered || !canExpand) return
+		const handler = (e: KeyboardEvent) => {
```
Makisuo: Please use https://tanstack.com/hotkeys/latest/docs/overview for hotkey managment ratehr than this manual one

apps/web/src/components/dashboard-builder/widgets/widget-shell.tsx
```diff
@@ -81,8 +82,37 @@ export function WidgetShell({
...
+		}
+		window.addEventListener("keydown", handler, { capture: true })
+		return () => window.removeEventListener("keydown", handler, { capture: true })
+	}, [hovered, canExpand])
+
 	return (
-		<Card className="group/card h-full flex flex-col">
+		<Card
+			className="group/card h-full flex flex-col"
+			onMouseEnter={() => setHovered(true)}
```
Makisuo, line 117: should use real focus states rather than this hovered state i think this is a bit hacky since this wont work with focus rn

apps/web/src/hooks/use-time-range-keyboard.ts
```diff
@@ -0,0 +1,169 @@
...
+}: UseTimeRangeKeyboardControls): void {
+	// Read the latest range/handler from a ref so the capture listener stays
+	// attached across range changes instead of re-binding on every keypress.
+	const stateRef = useRef({ start, end, onChange })
+	stateRef.current = { start, end, onChange }
+
+	useEffect(() => {
+		if (!enabled) return
+
+		const handler = (e: KeyboardEvent) => {
```
Makisuo: same again use https://tanstack.com/hotkeys/latest/docs/overview

Makisuo (changes requested):

Makisuo:
@pullfrog fix all issues

PR #110, opened 2026-06-21
feat(web): correlate traces/logs to their infrastructure

Makisuo:
## What

Opening any span or log that carries Kubernetes/host resource attributes now surfaces an **"Infrastructure"** tab in the detail panel. It shows that pod / node / host's live CPU / memory / disk charts and a deep-link into the existing infra detail pages.

The tab appears on:
- the **trace span detail panel**
- the **log detail sheet**

…and only when the record actually carries infra metadata (`k8s.pod.name`, `k8s.node.name`, `host.name`).

## Why

This was inspired by studying how **HyperDX** integrates infrastructure with logs/traces (its `infraCorrelations` / `DBInfraPanel` pattern). Maple already had a complete infra stack — pods/nodes/workloads + host pages, single-entity queries, cheap API endpoints, and reusable chart components — but no bridge *from a signal back to its infrastructure*. This closes that gap.

> Deliberately **not** porting HyperDX's flexible "Sources" field-mapping abstraction — Maple has a fixed Tinybird/ClickHouse schema, so that indirection buys nothing.

## How

**No backend changes.** Purely additive frontend that reuses the existing single-entity infra endpoints and chart components.

- `apps/web/src/components/infra/infra-correlations.ts` — pure, React-free detector (`getActiveInfraCorrelations`). Single source of truth for both the tab-visibility gate and the panel renderer, so they never drift. Discriminated union on `kind` keeps each group's metric-name literals exact (no inline casts).
- `apps/web/src/components/infra/infra-correlation-panel.tsx` — Maple's analogue of HyperDX's `DBInfraPanel`. Renders `PodDetailChart` / `NodeDetailChart` / `MetricStrip` over a window padded around the span/log timestamp, with typed SPA deep-links. Thin — the chart components self-fetch and own their loading/error/empty states.
- Wired a conditional Infrastructure `TabsTrigger` + `TabsContent` into `span-detail-panel.tsx` and `log-detail-sheet.tsx`.

## Verification

- **Unit tests** — 9 pass (`infra-correlations.test.ts`): pod/node/host/pod+node/empty detection, empty-string guarding, namespace query param, URL-encoding.
- **Typecheck** — `bun --filter=@maple/web typecheck` clean.
- **Live browser (log detail sheet, real dev-org data)** — a `subscriptions-api` log on Fargate surfaced all three groups with correct deep-links (pod incl. `?namespace=default`, node, host) and **12 charts rendered with real data**; tab correctly absent on records without infra metadata.

### Reviewer notes
- The span detail panel reads the lazily-loaded full resource map (`detailAttrs.resourceAttributes`) and falls back to the trimmed `span.resourceAttributes` so the tab can appear before the detail query resolves.
- The span surface shares the exact same panel + detector as the (live-verified) log surface; its wiring is typecheck-clean and structurally identical, but I could not get a fresh live span click after the preview restarted into a billing-gated org state where `/traces` bounces to the subscription wall (environmental, not a code issue).

### DIFF
--- a/apps/mobile/app/(home)/settings.tsx
+++ b/apps/mobile/app/(home)/settings.tsx
@@ -1,12 +1,12 @@
 import { useState } from "react"
-import { useClerk, useOrganization, useUser, useUserProfileModal } from "@clerk/expo"
+import { useClerk, useOrganization, useUser } from "@clerk/expo"
 import { UserButton } from "@clerk/expo/native"
 import { Alert, Linking, Pressable, Text, View } from "react-native"
 import { Screen } from "../../components/ui/screen"
 import { ScreenHeader } from "../../components/ui/screen-header"
 import { SectionHeader } from "../../components/ui/section-header"
 import { Card } from "../../components/ui/card"
-import { DestructiveButton, SecondaryButton } from "../../components/ui/button"
+import { DestructiveButton } from "../../components/ui/button"
 import { OrgSwitcherModal } from "../../components/org-switcher-modal"
 import { hapticWarning } from "../../lib/haptics"
 import { colors } from "../../lib/theme"
@@ -14,7 +14,6 @@ import { colors } from "../../lib/theme"
 export default function SettingsScreen() {
 	const { signOut } = useClerk()
 	const { user } = useUser()
-	const { presentUserProfile } = useUserProfileModal()
 	const { organization } = useOrganization()
 	const [orgModalVisible, setOrgModalVisible] = useState(false)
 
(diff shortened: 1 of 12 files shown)


Makisuo:
Addressed the `href` rough edge in `cadd96502`: dropped the dead `href` field from `InfraCorrelation` and all three builders. `CorrelationLink` already navigates via TanStack `<Link to/params/search>` (which owns URL encoding), so `href` was never read. The three `href` assertions are replaced with ones that check the raw `identifier`/`namespace` actually driving the `<Link>` — so coverage now guards the shipped navigation path, not an unused string. Typecheck clean, 39 infra tests green.

PR #114, opened 2026-06-24
feat(traces): surface cloud-platform span annotations in trace views

Makisuo:
## What

Surfaces serverless/cloud-platform span annotations across all three trace views (waterfall, flow diagram, span detail panel). Cloudflare Worker spans — e.g. `maple-chat-flue` — come from Cloudflare's native **Workers Observability** pipeline (`scope.name: workers-observability`), so they carry a rich attribute set no other emitter sends, none of which the trace views previously displayed:

- **Edge location** — `cloudflare.colo`, `faas.invoked_region`, `geo.*`
- **Worker outcome** — `cloudflare.outcome` (ok / exception / exceededCpu / canceled), distinct from the OTEL span status
- **Performance** — `cpu_time_ms` vs `wall_time_ms`, TTFB
- **Identity / cross-ref** — script name + version, execution model, handler, `cloudflare.ray_id`

## Why

These annotations sat inert in `SpanAttributes`, only visible in the raw attributes table. Surfacing them makes Worker traces legible at a glance (where it ran, whether the isolate succeeded, CPU pressure) and lets you cross-reference the Ray ID against the Cloudflare dashboard.

## How — built extensible for other providers

Implemented as a **platform-adapter registry** at `packages/ui/src/lib/cloud-platforms/`, so adding Vercel Functions / AWS Lambda later is a new adapter file + one line in `ADAPTERS` with **zero component changes**. The trace UI consumes only a normalized `CloudPlatformInfo` (`Icon`, `label`, `kind`, `edge`, `location`, `outcome.{value,bad}`, `fields[]`):

- `types.ts` — normalized shape + shared helpers (`outcomeBadgeStyle`, `pickAttr`, `joinEdge`)
- `cloudflare.ts` — the `cloudflareAdapter` (only place that knows `cloudflare.*` keys)
- `index.ts` — `getCloudPlatform(attrs)` registry + documented extension point

Consumers:
- **span-row** — provider mark on the service badge, edge chip, bad-outcome badge
- **flow-node** — provider icon + kind label in the header, edge/outcome in the footer
- **span-detail-panel** — provider summary section that iterates `fields[]` (copyable Ray ID); the row component is provider-agnostic

## Reviewer notes

- **`TREE_SPAN_ATTR_KEYS` (query-engine `errors.ts`)** — the waterfall/flow views read a *trimmed* attribute set; the relevant `cloudflare.*` / `cloud.platform` / `faas.invoked_region` keys were added there so the tree-view badges aren't blank (selecting the full `SpanAttributes` map per span OOMs wide traces; the detail panel lazy-loads the full map and is exempt).
- **Detection precision** — `buildProjectedMapExpr` emits each requested key with an *empty-string value* on every span, so detection keys on the exact `cloud.platform === "cloudflare.workers"` signal and only treats `cloudflare.*` as a fallback when it has a **non-empty value**. A key-presence check would (and initially did) flag every span as a Worker; there's a regression test for this.
- Added a `CloudflareIcon` to `packages/ui` (it only had `GlobeIcon`); explicit `"./lib/cloud-platforms"` export entry in `packages/ui/package.json`.

## Verification

- **Tests** — `cloud-platforms.test.ts` 8/8 (incl. the empty-projected-keys regression + a precise-detection case where a `via: cloudflare` header value must *not* trip detection); query-engine `errors.test` green.
- **Typecheck** — clean for `packages/ui` and `apps/web`.
- **Browser** — waterfall + 33-node flow + detail panel render cleanly; non-CF spans show no platform badges/section (the dev login org has no Workers Observability traces, so live CF badges weren't visually confirmed — covered by unit tests + a captured-fixture test).


Makisuo:
Addressed the combined-node note (a0cb2566): gated the edge branch on `!isCombined` in `flow-node.tsx`, so combined Worker nodes now show the `{count} spans` aggregate (the more useful signal when aggregated) while single Worker nodes still show the edge label.

PR #120, opened 2026-06-25
perf(services): batch service-detail queries + lengthen org CH config cache

Makisuo:
## Why

The service-detail page is the worst user-facing latency in the app. Live telemetry (Maple MCP, prod) showed `maple-web` p95 at **3.56s**, with the client-perceived `serviceOverview` atom running **3.6–4.1s p95** while the underlying SQL was only **~0.8s**. That 3× gap — not slow SQL — was the target.

Two root causes accounted for nearly all of it:

1. **Cold org-config resolution dominated the tail.** Every warehouse execute resolves per-org ClickHouse config; on a memo + edge-cache miss it does a Postgres/Hyperdrive handshake measured at **0.85–2.4s** in prod. `0.8s SQL + ~2.4s cold config ≈ 3.2s`, which matches the observed client p95.
2. **The Overview tab fanned out 3 separate Worker requests** — each re-resolving config and paying its own browser→Worker round-trip. One of them was a wasteful **all-services** overview scan fired only to populate the environment-switcher dropdown.

(All the heavy service-detail queries already have `Timestamp`/`Hour` partition-pruning predicates — the bottleneck was fan-out + the cold-config tail, not the SQL.)

## What changed

**1. Org-config cache TTL bump** (`apps/api/src/services/OrgClickHouseSettingsService.ts`)
- In-isolate memo: 30s → **5 min**; edge cache (KV): 300s → **1 h**.
- Config is invalidated on every upsert/delete, so longer TTLs lose no correctness — they just cut cold Postgres handshakes ~12×. This directly attacks the ~2.4s tail.

**2. Overview-tab bundle (3 requests → 1)** — new `serviceDetailOverview` endpoint
- Runs the primary all-metrics chart, the releases timeline, and a new **service-scoped** distinct-environments query (`serviceEnvironmentsQuery`) under a single tenant/config resolution.
- The new environments query replaces the all-services overview scan; it's service-scoped + partition-pruned.
- `OverviewTab` and `ServiceEnvironmentSwitcher` read the **same** atom key, so `Atom.family` dedupes them to one fetch.

**3. Dependencies-tab bundle (3 requests → 1)** — new `serviceDependenciesBundle` endpoint composing the service / DB / external edge queries in one request.

## Verification

- **Typecheck**: green across all 23 packages (the API client is auto-derived from the HttpApi types).
- **Tests**: query-engine (624), api (607), web (467) all pass. Added a unit test for `serviceEnvironmentsQuery` asserting the org + service + time-window predicates.
- **End-to-end against the real warehouse** (authenticated call from the browser session):
  - `service-detail-overview` → **200**, `{ timeseries: 48 pts, releases: 2, environments: ["production"] }`.
  - `service-dependencies-bundle` → **200**, `{ dependencies, dbEdges, externalEdges: 2 }`.
- The live page render couldn't be exercised because the test org is behind a "PICK A PLAN" billing wall that redirects `/services/*` — so the endpoints were verified directly instead.

## Reviewer notes

- The request consolidation (3→1) is structurally guaranteed: both the chart grid and the env switcher read one atom keyed identically, deduped by `Atom.family`.
- Expected impact: the slowest service-detail query (the all-services `serviceOverview` scan, 812ms SQL / 3.6s client) is gone, and combined with the config TTL bump the cold-path tail should drop the client p95 from ~3.6s toward ~1s.
- The rewiring orphaned five now-unused exported atoms (e.g. `getCustomChartServiceDetailResultAtom`, `getServiceMapForServiceResultAtom`); they tree-shake out of the build and were left in place to keep the diff focused. Easy follow-up to delete.


Makisuo:
Thanks for the review. Both notes were informational (no required change), but I addressed the concrete one:

**Orphaned atoms (ℹ️ #2)** — removed all five now-unused atoms (`getCustomChartServiceDetailResultAtom`, `getServiceReleasesTimelineResultAtom`, `getServiceMapForServiceResultAtom`, `getServiceMapDbEdgesForServiceResultAtom`, `getServiceExternalEdgesResultAtom`) and their now-unused imports in `warehouse-query-atoms.ts` (commit e9e80a7). The underlying fetch functions stay as module API — `getCustomChartServiceDetail` still has unit coverage, and the others are exported utilities that tree-shake; removing those too would cascade into multi-file import pruning best left to a dedicated cleanup.

**Default-window divergence (ℹ️ #1)** — left as-is intentionally. The branch is dead in the current call graph (the route always passes `effectiveStartTime`/`effectiveEndTime`), and where it differs, the bundle's explicit 24h default is the more sensible behavior than the standalone's wide-open `2020→2099` span + fixed 300s bucket. `getCustomChartServiceDetail` (the standalone path) is kept and still test-covered, so I didn't want to delete it to force convergence.

Typecheck green across all packages; web suite (467) still passing.

PR #123, opened 2026-06-25
fix(billing): short-cache planless customers to end post-checkout onboarding lockout

JeremyFunk:
## Problem

On the PR-preview deploy, a user who completed Stripe checkout (test card) was **permanently bounced to `/quick-start` onboarding** on every navigation, despite having a valid trial.

The root cause is a stale **negative** cache, not the checkout itself. After checkout, Autumn records the new subscription asynchronously. A customer with **no active plan** was cached for the full 5 minutes at two layers — the API per-org edge cache (`getOrCreateCustomer`) and the client react-query layer. The post-checkout return is a fresh page load, and the `attach` mutation that invalidates the cache already ran *before* the redirect to Stripe, so nothing re-invalidates on return. If the first post-return `getOrCreateCustomer` landed before Autumn finished syncing, the "no plan" snapshot stuck for 5 minutes and the `__root.tsx` plan gate re-bounced the user on every navigation.

This is pre-existing behavior, unrelated to the recharts upgrade, and it affects **real paying customers** — not just preview.

## Fix

Treat a planless customer as an *unsettled* state that's cheap to re-check, at both layers:

- **API edge cache** (`autumn.http.ts`): `getOrCompute` now accepts a content-dependent TTL. A confirmed active-plan customer keeps the full 300s; a planless or error-shaped `200` is cached for just **5s** (`CUSTOMER_CACHE_UNSETTLED_TTL_SECONDS`).
- **Client** (`use-maple-customer.ts`): dynamic `staleTime` (5 min when settled, 5s when planless) plus a self-stopping `refetchInterval` while planless, so the gate auto-recovers within seconds with no manual refresh. Polling stops the moment a plan is active and pauses while the tab is backgrounded.

## Avoiding drift

The gate rule ("is this an actively selected paid plan?") would now live in two places — the web redirect gate and the API cache TTL. Extracted it into a single source of truth, `@maple/domain/billing` `isActivePlanSubscription`, structurally typed so `autumn-js` stays out of `@maple/domain`. Both `plan-gating.ts` (web) and `autumn.http.ts` (api) delegate to it.

## Verification

- `bun typecheck` — 23/23
- Tests: domain billing (3), web plan-gating (19), api autumn (11), query-engine edge-cache (8)


JeremyFunk:
@pullfrog pls merge main into this

Makisuo (approved):

PR #127, opened 2026-06-25
chore: adopt mise-en-place for toolchain + CI parity

Makisuo:
## What

Adopt [mise-en-place](https://mise.en.dev) as the single source of truth for this repo's toolchain, and wire it into CI so **local dev and CI run identical tool versions**.

Today there's no version manager: the only pin is `packageManager: bun@1.3.11` (already drifting — dev machines run 1.3.14), Rust floats on `stable`, node is unpinned, and Python falls back to system 3.9.6. CI installs each runtime through a different action with no shared pins.

## Changes

- **`mise.toml`** (new) pins `bun 1.3.11` (matches `packageManager`), `node 24`, `rust 1.94.1` (apps/ingest; was a floating `stable`), `python 3.12` (Tinybird tooling + the CI benchmark script; up from 3.9.6).
- **`[env]`** auto-loads `.env.local` into the shell. **Additive** — apps still load it explicitly (portless `--env-file`, ingest `source`, scraper `--env-file`), so non-mise users and CI are unaffected; a missing file is a no-op (verified).
- **`[tasks]`** adds a `setup` bootstrap (tools + deps + env file + portless CA) plus thin `dev`/`build`/`test`/`typecheck`/`db:*` aliases that delegate to the existing `bun`/`turbo` scripts — `package.json` stays the source of truth.
- **CI**: replaced 12 per-tool setup steps across 10 workflows with `jdx/mise-action` (SHA-pinned `v2.4.4`) — `setup-bun` in 8 bun workflows, `dtolnay/rust-toolchain@stable` in the 2 ingest Rust workflows (kept `Swatinem/rust-cache`).
- **`.gitignore`**: ignore `mise.local.toml` (per-dev overrides).
- **`CLAUDE.md`**: document the mise quick-start.

## For reviewers

- **mise is optional locally** — `bun dev` / `bun test` / etc. still work without it. You only get pinned versions + env auto-load if you `mise activate` in your shell.
- **`bun` must stay in sync with `packageManager`** in `package.json` (noted in `mise.toml` + CLAUDE.md).
- **CI trade-off**: with one root config, bun-only jobs also install rust/python — modest, and cached by `mise-action` after the first run. If JS-job time becomes a concern, rust can move to `apps/ingest/mise.toml` later.

## Verification

- `mise install` → bun **1.3.11**, node **24.18.0**, rustc **1.94.1**, python **3.12.13**
- `mise run typecheck` → **23/23** turbo tasks pass via the mise-resolved bun
- `cd apps/ingest && mise exec -- cargo check` → clean under pinned rust
- `.env.local` auto-load confirmed; missing-file case is a graceful no-op

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
         steps:
             - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
 
-            - uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2
+            - uses: jdx/mise-action@c37c93293d6b742fc901e1406b8f764f6fb19dac # v2.4.4
 
             - run: bun install --frozen-lockfile
 
@@ -32,7 +32,7 @@ jobs:
         steps:
             - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
 
-            - uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2
+            - uses: jdx/mise-action@c37c93293d6b742fc901e1406b8f764f6fb19dac # v2.4.4
 
             - run: bun install --frozen-lockfile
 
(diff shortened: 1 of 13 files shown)


Makisuo:
Addressed in aa3309972 — pinned `node = "24.18.0"` and `python = "3.12.13"` (the exact patches from the verification block), matching the already-exact bun/rust. The whole `[tools]` table is now exact, so `mise install` resolves byte-identical toolchains local↔CI with no `mise.lock` needed. And yes — patch-level parity was the intent ("identical tool versions"), so this tightens it to the stronger guarantee.

(Noting the bot's side comment about the SHA-pinned action freezing its cleanup step: kept the SHA pin to match this repo's convention for third-party actions — a Dependabot bump is the repo-wide answer there, out of scope for this PR.)

PR #140, opened 2026-06-27
perf(traces): probe trace timestamp to prune partitions in spanHierarchy

Makisuo:
## What

The trace-detail page's blocking query — `QueryEngine.getSpanHierarchy` — was running **p95 ~8.8s on ~41% of loads**. This adds a cheap server-side timestamp probe so every load prunes ClickHouse partitions and hits the fast path (~2.3s).

## Why

`trace_detail_spans` is `PARTITION BY toDate(Timestamp)` (30-day TTL), sort key `(OrgId, TraceId, SpanId)`. Filtering `OrgId + TraceId` is a fast index seek **within a partition**, but with no `Timestamp` predicate ClickHouse can't prune and must seek + project across all ~30 daily partitions (it also `ORDER BY Timestamp ASC LIMIT 5000` and JSON-stringifies two maps per row).

The query only narrows when the client sends a window, which it only has when the `t` timestamp search param is in the URL. Most in-app links pass `t`, but ~41% of loads arrive without it: direct URLs, shared links, refresh-from-no-`t`, and AI surfaces (`ai-triage-card`, `inline-trace`) whose data carries only a trace id.

Measured over 7 days (prod, all orgs), split by whether the SQL has a `Timestamp` WHERE predicate:

| Query shape | n | p50 | **p95** | max |
|---|---|---|---|---|
| **No time filter** (full scan) | 46 (**41%**) | 1003ms | **8823ms** | 10816ms |
| **With time filter** (1 partition) | 65 | 444ms | **2303ms** | 3160ms |

## How

Entirely server-side — no client, URL, or response-schema changes:

- **`packages/query-engine`** — new `traceTimeProbeQuery`: a `LIMIT 1`, no-`ORDER BY`, single-column (`Timestamp`) seek that early-terminates per partition to resolve any one span timestamp for a trace. Exported from `ch/index.ts`.
- **`apps/api` (`query-engine.http.ts`)** — when the `spanHierarchy` handler receives no time window, it runs the probe (span context `spanHierarchyProbe`), derives a ±1h window, and compiles the hierarchy query narrowed. If the probe finds nothing (trace outside retention) it falls back to the prior unbounded compile, so that edge case is unchanged. Renamed the generic `logWindowAround` helper to `partitionWindowAround` and reused it.
- **Test** — locks in the probe's cheap shape (`LIMIT 1`, no `ORDER BY`, no `toJSONString`).

The lazy span-detail/logs queries already narrow via `span.startTime`, so they were never the page-load blocker and are untouched.

## Reviewer notes

- The performance win only manifests against production-scale partitioned data — it can't be reproduced on local dev (sparse trace data, no partition pressure). Correctness is covered by the compile test + typecheck; verify the win post-deploy by re-running the `run_sql` split above and confirming the no-time-filter bucket collapses toward zero, with a sub-100ms `spanHierarchyProbe` span.
- Diagnosed via Maple's own MCP against production telemetry.

## Verification

- `bun run --cwd packages/query-engine test` — 154 pass (incl. new probe test)
- `bun typecheck` — 24/24 packages green


Makisuo:
Thanks for the review. Addressed the non-blocking observation in 559d6d0fa: moved the timestamp probe inside `cachedDirect`'s effect, so it now runs only on a cache miss (Effects are lazy, so `edgeCache.getOrCompute` won't fire it on a hit). No behavior change otherwise — tests (154) and `@maple/api` typecheck green.

PR #194, opened 2026-07-09
Local telemetry archives for Maple's embedded store (depends on #129)

robbiemu:
# Local telemetry archives for Maple's embedded store

> **Depends on PR #129.** Archives use the validated checkpoint and restore
> primitives provided by that change and must not merge before the dependency
> is available.

## Summary

This PR adds long-term, portable Parquet telemetry archives to Maple's local
mode. It exports the six raw telemetry tables (`logs`, `traces`, `metrics_sum`,
`metrics_gauge`, `metrics_histogram`, and `metrics_exponential_histogram`) from
validated checkpoints into immutable UTC-day generations that can be queried
directly with DuckDB.

Archives do not reload history into the live store, add a live export endpoint,
or require a second always-on database.

## What changed

- `archive create` exports one signal and UTC day from restored checkpoint
  scratch storage and seals a validated Parquet generation.
- `archive list` exposes active generations or verified shard paths for DuckDB;
  `archive rebuild` reconstructs catalogs from manifests.
- Late telemetry creates a newer generation selected by an atomic active
  pointer. Superseded generations remain retained until safe GC.
- `archive gc` uses ownership-aware tombstone retirement and crash recovery.
- `archive calibrate` measures the deployment's real export behavior and can
  write a hash-bound tuning configuration.
- Every mutation is journaled and reconciled after interruption. Export never
  opens or mutates the live chDB store.

## Safety model

Export restores one validated checkpoint into private sacrificial scratch
storage. A persistent checkpoint pin is acquired under the maintenance lock,
protecting the source from retention while export runs. Checkpoint operations,
export, calibration, reset, restore, and GC serialize through that lock.

Generations are immutable. A late-arrival re-export supersedes an older
generation structurally rather than attempting unsafe cross-table `TraceId`
deduplication. Manifests bind source checkpoint identity, shard hashes, row
counts, event-time bounds, schema identity, and effective tuning.

## Calibration compatibility

New calibration documents are `formatVersion: 3` and use directional resource
cost evidence: lower observed RSS, wall time, compression ratio, physical bytes,
or temporary-disk peak is safe; only a regression beyond tolerance fails.

The loader remains compatible with immutable `formatVersion: 2` operator
artifacts. A v2 document is accepted only if its complete held-out evidence
matches one coherent historical policy:

- the original symmetric evidence representation, or
- the transitional directional representation emitted during the Phase 3B
  repair.

Format 3 accepts directional evidence only. A document mixing policies across
selected evidence or held-out attempts is rejected. Manifest generation format
3 preserves the loaded tuning identity, including v2 and v3 config identities.

Calibration child processes write `/usr/bin/time` output to a private per-child
report and wait for `close` so worker stdout/stderr is fully drained before the
next candidate starts.

## Operational workflow

1. Ingest telemetry into Maple.
2. Create a validated checkpoint.
3. Optionally run `archive calibrate <day> --write-config <path>`.
4. Run `archive create <day> <signal> --config <path>` for each signal.
5. Run `archive verify` to stream and verify active shard contents.
6. Use `archive list --output paths` to obtain the verified active Parquet paths.
7. Query them with DuckDB using `read_parquet(..., union_by_name=true)`.

The source row count, archived row count, and sum of shard row counts must
agree before a generation is published.

## Validation evidence

Source validation for the current hardening commit:

- 324 CLI tests across 18 files: pass.
- CLI typecheck: pass.
- Formatting and `git diff --check`: pass.
- Independent review: no blocking or non-blocking findings.
- The captured Phase 3B v2 tuning artifact (SHA-256
  `5d06ba124c36304754aed2c0960a8a17aad6db762b078ce6b13a15f3a8b1c88b`) loads
  unchanged with format-2 identity and calibrated overrides. Its 17 lower-cost
  resource comparisons all use directional `relativeDelta: 0` evidence.

Previously deployed Phase 3B validation on the r4 predecessor passed the
archive lifecycle, calibration, config-bound creation, and DuckDB verification.
The DuckDB check read generation `17d4f48a` across two shards and found both
marker rows. A later config-bound archive sealed generation
`f3166830-299f-4396-8c04-619e9b201ffc` with 35,112 log rows across four shards
and a hash-bound tuning identity.

## Deployment status

The deployed r4 image remains the proven Phase 3B artifact. Commit `269053b4`
is source-only compatibility and timing hardening; it has not yet been built
or deployed. A fresh image build and native probe are required before claiming
the current commit itself is deployment-proven.

## Known limits and follow-up

- Rotation remains operator-initiated; scheduling should follow measured
  checkpoint-pause characterization.
- Archives do not prune the hot store or rehydrate historical data into Maple's
  UI.
- Calibration is machine-, schema-, and archive-volume-specific.
- There is no always-running archive database; DuckDB opens files on demand.

### Devin review follow-up — 2026-07-10

- Fixed the calibration no-recommendation path so it fails through Effect’s typed `ArchiveError` channel rather than throwing inside `Effect.sync`.
- Consolidated the command on the shared archive error type and mapped calibration config write failures into the typed error channel.
- Added regression coverage for both missing and successful recommendations.
- Validation: 326/326 CLI tests, CLI typecheck, formatting, lint, and `git diff --check` pass.
- Fix commit: `1378ecce`.

### Review follow-up and checkpoint restore stabilization 2026-07-12

The archive review exposed an intermittent failure while reopening checkpoints through chDB’s scratch `RESTORE DATABASE` path. We reproduced it independently in the checkpoint implementation, addressed it there first, and then merged the fix into this branch.

Investigation showed that chDB v26.1.0 uses a separate 16-thread restore pool even when table loading is serialized. Timing delays and database-readiness probes did not reliably eliminate the failure, while serializing restore workers with `--restore_threads=1` survived the expanded checkpoint and archive stress tests.

This deliberately trades restore parallelism for reliability in Maple’s embedded workload. The 16-thread value is chDB/ClickHouse’s normal default, and the precise upstream mutex defect remains unidentified; the evidence does not establish that multithreaded restore is universally unsafe.

#### Combined validation note

At the current-main-merged #194 head, validation completed successfully:

- 341/341 CLI tests
- CLI typecheck
- exact-head native archive lifecycle across all six signals, including explicit integrity verification, DuckDB counts, live-store invariance, and catalog rebuild
- 18/18 native crash-recovery cases: all 17 SIGKILL boundaries plus reconcile-before-new-create
- the pre-main-merge stress cohort passed 12/12 complete archive lifecycles (72 restored signal exports)
- the underlying checkpoint fix separately passed 300/300 focused restores and 10/10 complete native checkpoint lifecycles; its current-main merge also passed a complete native lifecycle
- archive lifecycle and crash-recovery probes are now required pull-request CI jobs, with a checksum-pinned DuckDB CLI and failure-evidence upload
- workflow YAML parsing and native-script syntax checks
- `git diff --check`

All seven GitHub review threads have replies and are resolved. The review fixes, updated checkpoint dependency, current `main`, and required native archive/recovery CI are committed locally and ready for the planned combined push.


apps/cli/src/commands/archive.ts
```diff
@@ -0,0 +1,1667 @@
...
+					)
+				}
+				const writePath = Option.getOrUndefined(a.writeConfig)
+				if (writePath) {
+					if (rec.selected === null) {
+						// No config emitted: a held-out pass did not succeed across the
+						// required signals, or no candidate met the declared goals.
+						throw new ArchiveError({
+							message: `${red("!")} calibration did not produce a recommendation: ${rec.note}`,
+						})
```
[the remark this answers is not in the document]
robbiemu: Fixed in 1378ecce and 8330749a. Calibration no-recommendation now fails through Effect’s typed `ArchiveError` channel outside `Effect.sync`, config publication uses `Effect.try`, and the CLI handles expected archive errors without diagnostic stack frames. Verified with 327/327 CLI tests, typecheck, formatting, lint, and a live deployment smoke on maple vfork-8330749a: nonzero clean error, no config emitted, no scratch debris, my local Maple healthy, and other containers unaffected.

apps/cli/src/server/checkpoints.ts
```diff
@@ -0,0 +1,1893 @@
...
+				await syncTree(join(restoreData, "backups"))
+				await durableJson(restoreReadyPath(dataDir, operationId), {
+					formatVersion: 1,
+					operationId,
+					checkpointId: resolvedCheckpoint.checkpointId,
+				})
+				await syncTree(restoreData, { allowSymlinks: true })
+				transaction = { ...transaction, phase: "restore-ready", validation }
+				await writeRestoreTransaction(dataDir, transaction)
+				await reconcileRestoreTransactionUnlocked(dataDir)
```
Makisuo, line 2050: This validates the restore through the temporary chDB instance, then publishes it as the live store without proving a fresh process can reopen it. The committed native checkpoint smoke restores successfully but the next server start crashes while loading the restored metadata. We need to fix the restored representation/lifecycle and require a repeated create -> restore -> fresh-process reopen test before merging.

apps/cli/src/commands/archive.ts
```diff
@@ -0,0 +1,1675 @@
...
+			const checkpointId = Option.getOrUndefined(a.checkpointId) ?? "current"
+			const budget: CalibrationBudget = {
+				memoryBudget: a.memoryBudget,
+				timeBudget: a.timeBudget,
+				sampleRows: a.sampleRows,
+				maxCandidateWallMs: a.maxCandidateWallMs,
+				minThroughputBytesPerSec: a.minThroughput,
+				maxTempDiskBytes: a.maxTempDisk,
+				freeSpaceReserve: a.freeSpaceReserve,
+				safetyMargin: a.safetyMarginMilli / 1000,
```
Makisuo: This budget is built directly from unrestricted CLI integers. `--safety-margin-milli 0` makes the RSS/disk checks automatically pass and makes throughput effectively infinite; negative values also disable the cost ceilings. Please validate the whole budget before any pin/session/I/O: positive safe integers for the ceilings and sample sizes, non-negative throughput, and a finite safety margin of at least `1`.

apps/cli/src/commands/archive.ts
```diff
@@ -0,0 +1,1675 @@
...
+					ok: false,
+					error: timeOutput.error
+						? `${timeOutput.error} (fail-closed)`
+						: `failed to parse peak RSS from /usr/bin/time report (fail-closed)`,
+				})
+				return
+			}
+			try {
+				const lines = stdout.trim().split("\n")
+				const raw = JSON.parse(lines[lines.length - 1]!) as ChildMetrics
```
Makisuo: This cast trusts the child-process protocol without runtime validation. Missing, `null`, string, or otherwise invalid numeric fields can become zero/`NaN`, and the later `>`/`<` checks fail open for `NaN`. Parse as `unknown` and decode an exact Effect Schema with finite non-negative metrics and safe-integer counts before deriving or selecting anything.

apps/cli/src/commands/archive.ts
```diff
@@ -0,0 +1,1675 @@
...
+					sample.checkpointId !== checkpointId ||
+					typeof sample.checkpointManifestFingerprint !== "string" ||
+					sample.checkpointManifestFingerprint !== checkpointManifestFingerprint ||
+					typeof sample.rangeDate !== "string" ||
+					sample.rangeDate !== rangeDate ||
+					(sample.role !== "training" && sample.role !== "held-out") ||
+					typeof sample.startRow !== "number" ||
+					typeof sample.requestedRows !== "number" ||
+					typeof sample.rowCount !== "number" ||
+					sample.rowCount !== raw.rowCount
```
Makisuo: The parent only checks that these scope fields are numbers; it never binds them to the `startRow` and `sampleRows` passed to this child. A held-out run could report the training window and still pass live validation. Require exact equality for the requested start row, requested row count, expected role, and returned row count.

apps/cli/src/server/archives/generation.ts
```diff
@@ -0,0 +1,1348 @@
...
+		await reconcileArchiveGeneration(dataDir, archiveDir, tuning.scratchRoot, faults)
+		// Reject impossible capacity before checkpoint resolution, pointer/base
+		// reads, or creation of a new durable intent. A failed preflight must not
+		// leave an operation for reconciliation to clean up.
+		await preflightFreeSpace(
+			archiveDir,
+			tuning.archiveDir,
+			tuning.minFreeSpaceReserve,
+			estimatedWorkingBytes,
+		)
```
Makisuo, line 376: This only preflights the archive filesystem using `targetChunkBytes`, but the checkpoint is restored under `scratchRoot` and can be much larger. On separate volumes this can pass, create the durable intent/pin, then fail during restore. Resolve the checkpoint first and preflight archive + scratch independently using `manifest.backupBytes`, avoiding double-counting when they share a device.

apps/cli/src/server/archives/listing.ts
```diff
@@ -0,0 +1,327 @@
...
+		if (signalHasActive) signalsPresent.push(signal.name)
+	}
+	return { archiveDir, active, signals: signalsPresent, errors }
+}
+
+const messageOf = (error: unknown): string => (error instanceof Error ? error.message : String(error))
+
+/** Compute SHA-256 of a file. Reads the whole file into memory; acceptable for
+ *  bounded shards (maxShardBytes) but not suitable for unbounded files. */
+const sha256FileSync = (path: string): string => {
```
Makisuo: `archive list` calls this for every active shard, including summary output, so listing is O(total archived bytes) and allocates a full shard-sized buffer (256 MiB by default) per hash. Please stream the hash with bounded buffers and separate cheap metadata listing from explicit integrity verification.

Makisuo (changes requested):
Thanks for putting this together. It goes way further than I would have gotten myself with Maple Local, but from looking more into this, this for sure makes sense to add, looks extremely usefull. The state-machine and filesystem coverage here is strong, but I don't think this is safe to merge yet. The main blockers are the fresh-process checkpoint reopen failure and several calibration paths that currently fail open on invalid inputs or child output.

Please fix the inline findings, make the full create -> restore -> fresh-process reopen lifecycle repeatably green, and wire the native checkpoint/archive/recovery probes into required CI. This also needs to land after #129 and be brought up to date with main before merge.

robbiemu:
> Thanks for putting this together. ...This also needs to land after #129 and be brought up to date with main before merge.

glad to contribute, thank you for this amazing local Otel! is 129 on track to merge?

edit: ooh, I see, you've added a problem for that PR here. looking to reproduce

apps/cli/src/server/checkpoints.ts
robbiemu: Fixed in the #129 follow-up (`4fdcbb79`).

Your observation was correct: the restore path validated the checkpoint through a temporary chDB instance, then published it without proving that a fresh Maple process could reopen the restored store. I reproduced that exact gap against the pre-fix #129 head. Restore validation succeeded, but a later fresh process intermittently crashed while loading restored metadata or failed with `ASYNC_LOAD_WAIT_FAILED`.

The evidence points to a startup metadata-loader concurrency problem rather than a consistently invalid checkpoint representation: the never-restored control passed 20/20 opens, restored stores passed reliably with serial loader pools, and the failure was intermittent rather than permanent corruption.

The fix now validates the restored directory in a separate Maple process before publishing it, fails closed if that validation does not match, and makes chDB metadata loading deterministic with asynchronous database/system loading disabled and single-entry loader pools. The native checkpoint smoke now requires repeated fresh-process reopens after restore, dirty recovery, checkpoint creation, and reset/restore. The exact committed bundle passed the strengthened smoke and 100/100 additional fresh opens.

This closes the lifecycle gap identified here. The runtime fix belongs in #129; #194 itself did not contain the checkpoint reopen fix.

apps/cli/src/commands/archive.ts
robbiemu: Fixed in `6c25f76f`.

The complete operator-supplied calibration budget is now validated before root resolution, status output, checkpoint pinning, session creation, child processes, or filesystem I/O. Memory, total-time, per-candidate-time, sample-row, temporary-disk, and free-space values must be positive safe integers; minimum throughput must be a non-negative safe integer; and the safety multiplier must be finite and at least `1`.

This prevents zero, negative, unsafe, or non-finite values from weakening RSS, disk, wall-time, or throughput enforcement. Regression coverage exercises zero, negative, unsafe-integer, `NaN`, infinity, and sub-1 margin cases. The focused calibration suite, CLI typecheck, lint, formatting, and the full CLI suite pass.

apps/cli/src/commands/archive.ts
robbiemu: Fixed in `a7700b24`.

The parent now parses the child JSON as `unknown` and decodes it through an exact Effect Schema with excess-property rejection. Metric fields must be finite and non-negative, and row/scope counts must be non-negative safe integers. Missing, `null`, string-coerced, negative, unsafe, `NaN`, infinite, and unexpected fields all fail the candidate instead of reaching the ceiling comparisons.

The externally measured `/usr/bin/time` RSS remains authoritative, but the child-reported RSS field is also schema-validated so the protocol is exact rather than partially trusted. Regression tests cover the malformed protocol cases, and the focused and full CLI suites pass.

apps/cli/src/commands/archive.ts
robbiemu: Fixed in `a7700b24`.

After schema decoding, the parent now binds the child report to the exact request it sent: checkpoint ID, checkpoint manifest fingerprint, range date, expected training/held-out role, requested start row, and requested row count must all match exactly. The sample's returned row count must also equal the decoded metrics row count.

This prevents a held-out child from reporting the training window—or any other mismatched scope—while still being accepted for live validation. Regression tests independently alter the role, start row, requested rows, and returned row count and confirm that each mismatch fails closed. The focused calibration suite and full 334-test CLI suite pass.

apps/cli/src/server/archives/generation.ts
robbiemu: Fixed in `ec2fccd0`.

Archive creation now resolves and validates the selected checkpoint first so the preflight can use its immutable `manifest.backupBytes`. Before writing a new operation intent or acquiring its pin, Maple separately measures the archive and scratch filesystems using device ID plus filesystem type.

On separate devices, the archive check requires its configured reserve plus predicted Parquet working bytes, while the scratch check independently requires the checkpoint backup bytes. When archive and scratch share one device, those demands are combined into one check against that device's free space, so neither the required bytes nor the available capacity is double-counted.

Tests cover independent archive and scratch failures, exact same-device combined accounting, safe-integer overflow, and the invariant that a failed preflight creates no active operation journal. The focused generation suite, CLI typecheck, lint, formatting, and the full 336-test CLI suite pass.

apps/cli/src/server/archives/listing.ts
robbiemu: Fixed in `4948e809`.

`archive list` is now a cheap metadata operation: it validates active pointer and manifest topology, refuses symlinked/non-regular shards, and checks manifest-bound file sizes, but it no longer reads every shard's contents. JSON listing explicitly reports `integrity: "metadata-only"`, and summary output directs operators to the verification command.

Full content verification is now the explicit `maple archive verify` operation, optionally scoped with `--signal`. It processes shards sequentially and hashes each through a fixed 1 MiB stream buffer, so memory use is bounded independently of shard size. Any listing error, size mismatch, or SHA-256 mismatch fails the verification without reporting partial success.

Tests prove metadata listing does not hash same-size tampering, explicit verification accepts a shard spanning multiple stream buffers, and the same-size tampering is then rejected by SHA-256. Native archive, calibration, and merge probes now run explicit verification before handing active paths to DuckDB. The full 338-test CLI suite, typecheck, lint, formatting, shell syntax, and `git diff --check` pass.

robbiemu:
Ready for re-review. All requested review threads are addressed and resolved, the checkpoint fix landed on the dependency branch and is merged here, the branch is current with main, and checkpoint/archive/crash-recovery native probes are now wired into PR CI.
GitHub is currently holding three workflow runs for maintainer approval because this PR originates from a fork. Please approve those runs as part of the re-review so the new head’s checks can execute.

robbiemu:
fyi, my gitguardian flagged [92f3342](https://github.com/MapleTechLabs/maple/pull/194/commits/92f33425cce92daf5a1690cf1f14c980deae9a53) merge here. In packages/domain/src/http/v2/api-keys.ts at line 126. Looks safe, just through I should point out:

Yes, it is intentionally hard-coded—but as an OpenAPI documentation example, not as an operational API key.

Evidence:

- It appears inside `Schema.String.annotate({ examples: [...] })` and the example response object.
- It is surrounded by fictional documentation data such as `ci@acme.com`, a fake public ID, and example timestamps.
- Real Maple keys are generated using `randomBytes(32).toString("base64url")`, producing 43 characters after `maple_ak_`.
- The reported example has only 32 characters after the prefix, so Maple’s normal key-generation path could never produce it.
- Authentication hashes the supplied value and requires a matching HMAC hash in the `api_keys` database table. Merely matching the `maple_ak_` prefix provides no access.

Assessment: **intentional synthetic example; GitGuardian false positive; no credential rotation required.**

The only caveat would be if GitGuardian independently reports the key as verified “valid,” which would suggest someone manually inserted its hash into a deployment database. That is extremely unlikely given the deliberately fake format. Otherwise, close it as **test/example credential**.

Upstream could reduce future alerts by using something unmistakable like `maple_ak_<generated-secret>` or adding GitGuardian’s documented suppression marker, but the present string is not a leaked secret.

Makisuo:
Looks great now, love it!

PR #206, opened 2026-07-13
feat(alerts): add email as a notification destination

Makisuo:
## What

Adds **email** as a first-class alert notification destination alongside Slack, Discord, PagerDuty, webhooks, and Hazel. Recipients are **workspace members** — selected by user id and resolved to emails server-side — never free-form addresses.

- **Domain** (`packages/domain/src/http/alerts.ts`): `"email"` destination type, `EmailAlertDestinationConfig` (name + 1–10 `memberUserIds`) and its update schema, registered in both request unions. `AlertDestinationDocument` exposes `memberUserIds` so the edit dialog can pre-select current recipients.
- **Member resolution** (`apps/api/src/services/OrgMembersService.ts`, new): resolves member user ids to `{userId, email, name}` via the Clerk backend (`getOrganizationMembershipList`), failing on ids that aren't members of the org. Create/update re-resolve server-side, so clients can never route alerts to arbitrary addresses. Self-hosted (non-Clerk) deployments get a clear "requires Clerk" validation error.
- **Storage** (`AlertsService`, `AlertDestinationHydration`): the resolved member snapshot lives in the encrypted secret config; the public config carries `memberUserIds` plus a name-based summary (`Ops Team +2 more`) and the first email as channel label. No DB migration needed (`type` is free-form text).
- **Delivery** (`AlertDeliveryDispatch`): `dispatchDelivery` gains a required `deps: { sendEmail }` parameter (same injection style as `fetchFn`) plus an `email` branch that renders once and sends to each member from the stored snapshot — no Clerk on the delivery path. Both callers (AlertsService outbox/test path and NotificationDispatcher) adapt the existing `EmailService` (Cloudflare `EMAIL` binding); layers wired in both the api and alerting workers. "Send test" works through the same path.
- **Content**: new react-email template `@maple/email/alert-notification` in the Maple dark style — event banner, severity/signal/group/observed/window table, "Open in Maple" + "Ask Maple AI" buttons — built from the same formatting helpers as the Slack/Discord payloads so channels never drift. Rendered in `apps/api/src/services/alert-email.ts`.
- **UI**: Email provider tile (envelope icon) and a Clerk-backed member picker (checkbox list with avatars via `useOrganization` memberships), pre-selected when editing.

## Reviewer notes

- Custom notification templates (`AlertNotificationTemplate`) are deliberately **not** applied to email in v1 — rendering arbitrary user Markdown into HTML email safely is out of scope; email always uses the built-in format.
- Member emails are **snapshotted at save time**; a member changing their email or leaving the workspace doesn't retroactively update stored destinations until the destination is edited. Acceptable v1 trade-off to keep Clerk out of the delivery path.
- `alert-email.ts` and `AlertDeliveryDispatch.ts` have a deliberate module cycle (formatters one way, render helper the other); both sides only reference the other's exports inside function bodies, so it's safe under ESM live bindings.
- Without the `EMAIL` binding (plain `bun dev`), a test send fails cleanly with "Email not configured: EMAIL binding is missing".

## Testing

- `bun typecheck` green across domain/api/alerting/email/web.
- Full `@maple/api` vitest suite passes (803 tests), including: per-member dispatch, send-failure mapping to `AlertDeliveryError`, create with server-side member resolution, non-member rejection, update replace-vs-keep, and test-send via stubbed `EmailService`/`OrgMembersService`.
- Template eyeballed via a rendered HTML preview.
- Not yet click-tested in the browser dev environment — worth a quick manual pass: Alerts → Destinations → New → Email → pick members → Send test.


Makisuo:
Reworked per feedback: email destinations now target **workspace members** instead of free-form addresses.

- The create/update request takes `memberUserIds`; the server resolves them to emails via the Clerk backend (`getOrganizationMembershipList`) and rejects ids that aren't members of the org — clients can never route alerts to arbitrary addresses.
- The resolved `{userId, email, name}` snapshot is stored in the encrypted secret config, so the delivery path (alerting worker) needs no Clerk access.
- The dialog's recipients field is now a member picker (Clerk `useOrganization` memberships, checkbox list with avatars), pre-selected from `AlertDestinationDocument.memberUserIds` when editing. Self-hosted deployments without Clerk see an explanatory note.

PR #212, opened 2026-07-15
fix(planetscale): forward signed data-plane auth params to branch scrapes

Makisuo:
## What & why

PlanetScale branch-metrics scraping was returning **`403 invalid signature`** on every branch, silently flatlining ingest for 22h+ in production.

Root cause: **the metrics data plane (`metrics.psdb.cloud`) does not authenticate with the service token or the OAuth bearer** — it uses a **signed, expiring URL**. PlanetScale's `http_sd` discovery response mints per-branch `__param_sig` / `__param_exp` labels, and Prometheus convention promotes `__param_<name>` → `?<name>=` query params on the scrape URL. The `Authorization` header (token/bearer) only authenticates the *discovery listing* on `api.planetscale.com`.

`subTargetsFromGroup` stripped every `__`-prefixed label except `__scheme__`/`__metrics_path__`, so it **dropped `sig`/`exp`** and built an unsigned URL → 403 on every scrape. Reconnecting and rotating service tokens never helped because the token was never the data-plane credential — this was a misdiagnosis baked into the integration's design.

## The change

- `subTargetsFromGroup` promotes `__param_*` labels onto a new **`PlanetScaleSubTarget.signedUrl`** (base `url` + `?sig=&exp=`).
- Base `url` and `subTargetKey` stay **param-free** on purpose: the signature rotates each 10-min discovery refresh, so keeping it out of the identity keeps the scraper's per-branch fiber (and `instance` label) stable — only the server-side fetch picks up the fresh signature.
- `ScrapeTargetsService.scrapeForCollector` and `PlanetScaleConnectionService.probeDataPlaneScrape` now fetch `signedUrl`.
- Corrected the now-wrong "data plane rejects OAuth bearers / only accepts service tokens" comments.

## Verification

- New discovery test asserts `__param_*` → `signedUrl`, `url`/`subTargetKey` stay stable, and sig/exp never leak into metric labels.
- **48 api tests pass**, **typecheck clean**.
- The URL built via `URLSearchParams` is **byte-identical** to one proven to return `HTTP 200` with live branch metrics (base64 `%3D%3D` padding included).

## Reviewer notes

- **Deploy scope: `apps/api` only.** The scraper delegates the actual fetch to the API's prometheus-scrape proxy, and its `target.url` was already the base URL — no scraper redeploy.
- Existing enabled targets recover on the **next scrape after deploy — no reconnect, no service-token step.**
- **Follow-up (not in this PR):** this proves the hybrid "paste a service token" flow is now unnecessary — the OAuth bearer auths discovery and the data plane only needs the signed URL, so `probeDataPlaneScrape` will pass for bearer targets and pure-OAuth scraping works end-to-end. Ripping out the manual-token UI + `setMetricsToken` is a clean simplification.

---

<a href="https://app.devin.ai/review/mapletechlabs/maple/pull/212" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/apps/api/src/services/PlanetScaleConnectionService.ts
+++ b/apps/api/src/services/PlanetScaleConnectionService.ts
@@ -179,14 +179,14 @@ export class PlanetScaleConnectionService extends Context.Service<
 		const decodeHttpSd = Schema.decodeUnknownEffect(Schema.fromJsonString(HttpSdResponse))
 
 		/**
-		 * Whether the bearer can perform an ACTUAL branch-metrics scrape. The SD
-		 * endpoint on api.planetscale.com accepting the bearer proves nothing
-		 * about the data plane: the discovered hosts (metrics.psdb.cloud) only
-		 * document service-token auth and reject OAuth bearers with 403. Probe one
-		 * discovered endpoint with the bearer so `readMetricsEndpoints` reflects
-		 * scraping, not listing. Inconclusive outcomes (no branches discovered
-		 * yet, transport blip, undecodable payload) keep the control-plane answer
-		 * instead of pausing a possibly-working target.
+		 * Whether an ACTUAL branch-metrics scrape works. The SD endpoint on
+		 * api.planetscale.com accepting the credential proves nothing about the
+		 * data plane: metrics.psdb.cloud authenticates with the signed `?sig=&exp=`
+		 * URL params minted in the SD response, so we probe a discovered endpoint
+		 * via its `signedUrl` and let `readMetricsEndpoints` reflect scraping, not
+		 * listing. Inconclusive outcomes (no branches discovered yet, transport
+		 * blip, undecodable payload) keep the control-plane answer instead of
+		 * pausing a possibly-working target.
 		 */
 		const probeDataPlaneScrape = Effect.fn("PlanetScaleConnectionService.probeDataPlaneScrape")(
 			function* (organization: string, bearer: string) {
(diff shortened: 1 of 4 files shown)


Makisuo:
**Re: Devin Review — "External scraper receives unsigned URLs … causing all scrapes to fail with 403"**

This is a false positive — the standalone scraper never fetches `target.url` and never contacts the PlanetScale metrics data plane directly.

The scraper (`apps/scraper`) is a scheduler, not a fetcher. Its only outbound HTTP calls are to the API and the OTLP ingest (`ApiClient.ts` / `OtlpIngest.ts`). For each scrape it calls `GET /api/internal/prometheus-scrape?targetId=…&sub=…` (`apps/scraper/src/ApiClient.ts:97-103`), passing **only** the id + subTargetKey — not the URL. The API proxy then resolves `match.signedUrl` server-side in `scrapeForCollector` (`apps/api/src/services/ScrapeTargetsService.ts:863`, the exact line this PR changes) and performs the signed fetch there, behind SSRF validation + the discovery cache.

The unsigned `url` on `InternalScrapeTarget` is consumed only for the fiber fingerprint (`ScrapeScheduler.ts:120`) and the `instance` label host (`ScrapeScheduler.ts:225`, host only — query params are irrelevant to both). Keeping it param-free is **deliberate**: the `sig`/`exp` rotate every 10-minute discovery refresh, so folding them into the identity URL would restart every per-branch scrape fiber on each rotation. No change needed.

PR #219, opened 2026-07-17
feat(cli): configure local bind host

robbiemu:
This PR broadens Maple Local from localhost-only use to lightweight local-network deployment, while preserving localhost as the safe default.

## What changed

- add `maple start --host <address>` and `MAPLE_LOCAL_BIND_HOST` while preserving the loopback default
- distinguish bind, connection/probe, and advertised browser hosts, including IPv4/IPv6 wildcard handling
- forward the resolved listener settings to detached children, CLI queries, and checkpoint requests
- keep embedded LAN/reverse-proxied Local UI traffic same-origin while preserving the hosted `local.maple.dev` loopback flow
- restrict browser CORS to the advertised same-origin UI and configured hosted UI instead of using a wildcard, rejecting arbitrary and DNS-rebinding origins
- document LAN exposure, `--offline`, hosted UI markers, IPv6 URLs, and the unauthenticated local-mode trust boundary

## Why

Local mode previously hardcoded `127.0.0.1`, so container and trusted-LAN deployments could not opt into a reachable listener without patching Maple or adding a forwarding process. The first implementation enabled the listener but a wider audit found that bind addresses are not always valid connection targets, the embedded UI needed an explicit same-origin path, and wildcard CORS was too broad for a listener that may be exposed beyond loopback.

Loopback remains the safe default. Binding outside loopback exposes local mode's combined unauthenticated listener, including OTLP ingest, `/local/query`, `/health`, and the bundled UI. Deployments should use it only on a trusted network or behind an authentication/TLS boundary.

## Validation

- CLI suite: 344 passed, 0 failed, including live IPv4 wildcard, IPv6 loopback, and IPv6 wildcard listeners
- Local UI suite: 4 passed, 0 failed
- CLI and Local UI TypeScript checks passed
- Local UI production build passed
- formatter and linter passed for all touched TypeScript files
- detached-child, checkpoint URL, hosted UI, same-origin UI, CORS, malformed port, and DNS-rebinding cases covered
- combined homelab integration candidate reconstructed its pinned tree and passed 47 deployment contract tests

This remains a draft pending upstream CI and review.

## Latest commit guide

- [`daab8fa7`](https://github.com/MapleTechLabs/maple/pull/219/commits/daab8fa7ca3012be37acb560af6e740a74634406) is the focused fix that made the embedded Maple Local dashboard work from a LAN hostname. It keeps the LAN/reverse-proxied dashboard same-origin while preserving the hosted `local.maple.dev` loopback behavior.
- [`405e7120`](https://github.com/MapleTechLabs/maple/pull/219/commits/405e71203c8a0811a4a9a54c4c4ab93fa5d9b039) is the wider follow-up audit and hardening sweep. It centralizes bind/probe/advertised address handling; covers IPv4 and IPv6 wildcard listeners; carries the selected host through CLI and checkpoint paths; tightens CORS and DNS-rebinding behavior; preserves custom hosted-dashboard compatibility; and adds the expanded tests and documentation.
- [`0621c43`](https://github.com/MapleTechLabs/maple/pull/219/commits/0621c4365936446d802850a29591f5df76c74f9e) addresses Devin's follow-up: localhost aliases and the Vite proxy are accepted only when both sides are loopback, preserving the LAN DNS-rebinding guard.
- [`8ae92bef`](https://github.com/MapleTechLabs/maple/pull/219/commits/8ae92bef6a8049688f694bd0abf3c76819a1bf24) addresses maintainer review: allowed loopback browser origins now receive matching CORS headers, the `Host`-header dependency of the DNS-rebinding guard is documented, bracketed IPv6 hosts are normalized, and malformed `MAPLE_LOCAL_UI_URL` values produce a clear CLI error.


robbiemu:
Maintainability follow-up: the immediate localhost/Vite regression is fixed in `0621c43` while retaining the PR's DNS-rebinding guard for non-loopback hosts.

That said, this audit exposed that local mode owns a fairly bespoke set of origin parsing, CORS response, advertised-host, hosted-dashboard, proxy, and DNS-rebinding rules. The behavior is tested, but the machinery feels more custom than should be necessary for a local HTTP server. It would be worth a separate follow-up to evaluate replacing or consolidating it with a standard Bun-compatible CORS middleware/policy layer, provided that any replacement preserves the requirements covered here: exact hosted origins, same-origin LAN/reverse-proxy use, loopback aliases and dev-proxy ports, Private Network Access headers, and DNS-rebinding rejection.

apps/cli/src/server/serve.ts
```diff
@@ -24,33 +24,67 @@ export interface AssetResolver {
...
+	} catch {
+		return false
+	}
+	if (originUrl.origin === corsOrigin) return true
+	// Compare host (including port), not scheme: a TLS reverse proxy may preserve
+	// Host while forwarding to this HTTP listener. Restrict the hostname to the
+	// bind/connect/advertised set so a DNS-rebinding origin cannot claim itself as
+	// a same-origin embedded dashboard.
+	return originUrl.host === requestUrl.host && browserHosts.includes(originUrl.hostname)
+}
```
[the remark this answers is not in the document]
robbiemu: Resolved in '0621c43' (see comment below, on the main PR)

apps/cli/src/server/serve.ts
```diff
@@ -24,33 +25,71 @@ export interface AssetResolver {
...
+
+export const corsHeadersForOrigin = (
+	origin: string | null,
+	corsOrigin: string,
+): Readonly<Record<string, string>> | undefined =>
+	origin === corsOrigin
```
Makisuo: this only echoes the exact hosted origin, so cross-origin loopback ingest passes isBrowserOriginAllowed but gets no allow-origin header and the preflight dies. e.g. a browser app on localhost:3000 posting otlp to 127.0.0.1:4318 — that worked with the old wildcard. intended? if not id echo any origin that already passed the allow check (at least the loopback ones, theyre local anyway)

apps/cli/src/server/serve.ts
```diff
@@ -24,33 +25,71 @@ export interface AssetResolver {
...
+	if (originUrl.origin === corsOrigin) return true
+	// Loopback aliases and dev-proxy ports are equivalent local origins. Require
+	// both sides to be loopback so this exception cannot weaken LAN DNS-rebinding
+	// protection.
+	if (isLoopbackHostname(originUrl.hostname) && isLoopbackHostname(requestUrl.hostname)) return true
+	// Compare host (including port), not scheme: a TLS reverse proxy may preserve
+	// Host while forwarding to this HTTP listener. Restrict the hostname to the
+	// bind/connect/advertised set so a DNS-rebinding origin cannot claim itself as
+	// a same-origin embedded dashboard.
+	return originUrl.host === requestUrl.host && browserHosts.includes(originUrl.hostname)
```
Makisuo, line 65: worth a one line comment that requestUrl.host is the client Host header here — the whole rebind guard leans on that. bun does set it from Host but its load bearing and easy to break in a refactor

apps/cli/src/lib/local-address.ts
```diff
@@ -0,0 +1,51 @@
+export const DEFAULT_LOCAL_HOST = "127.0.0.1"
+export const DEFAULT_LOCAL_PORT = 4318
+
+export const resolveBindHost = (environmentValue: string | undefined): string =>
+	environmentValue?.trim() || DEFAULT_LOCAL_HOST
+
+/** Wildcard bind addresses are not connection targets. Use their matching
+ * loopback address for same-machine clients and readiness probes. */
+export const connectionHostForBindHost = (host: string): string =>
```
Makisuo: if someone passes --host [::] (with brackets) this wont match :: so it never maps to loopback and bun.serve gets [::] instead of ::. minor but maybe strip brackets or just document bare addresses only

apps/cli/src/lib/local-address.ts
```diff
@@ -0,0 +1,51 @@
...
+export const canonicalUrlHostname = (host: string): string => new URL(serverUrl(host, 80)).hostname
+
+export const hostedDashboardUrl = (baseUrl: string, port: number): string => {
+	const url = new URL(baseUrl)
+	url.searchParams.set("port", String(port))
+	url.searchParams.set("maple-local-api", "loopback")
+	return url.toString()
+}
+
+export const hostedUiOrigin = (baseUrl: string): string => new URL(baseUrl).origin
```
Makisuo: throws if MAPLE_LOCAL_UI_URL is malformed and blows up start as a defect. fail fast is fine but a real error msg would be nicer

Makisuo (commented):
nice, the address split + rebind guard looks really good. couple things, and we can get it merged

apps/cli/src/server/serve.ts
robbiemu: Fixed in 8ae92bef. Once an origin passes isBrowserOriginAllowed, we now echo that origin in Access-Control-Allow-Origin. This restores browser OTLP ingest across loopback aliases and ports without restoring wildcard CORS. Added coverage for localhost:3000 posting to 127.0.0.1:4318.

apps/cli/src/server/serve.ts
robbiemu: Added the requested comment in 8ae92bef, explicitly noting that Bun constructs requestUrl from the client Host header and that this comparison is the load-bearing DNS-rebinding check for non-loopback traffic.

apps/cli/src/lib/local-address.ts
robbiemu: Fixed in 8ae92bef. Host inputs are now normalized before binding, so [::] becomes :: for Bun.serve and maps to ::1 for probes and local connections. Added regression coverage for the bracketed form.

apps/cli/src/lib/local-address.ts
robbiemu: Fixed in 8ae92bef. Hosted UI URLs are now validated as absolute HTTP(S) URLs, and invalid MAPLE_LOCAL_UI_URL values surface as a clear typed ServerError instead of an unhandled defect. Added tests for malformed and non-HTTP URLs.

Makisuo:
Looking good now thanks @robbiemu. Really appreciate it!

PR #265, opened 2026-07-26
docs(alchemy-provider): use the Effect-native Worker API in the example

Makisuo:
Follow-up to #264. The example stack shipped the **async-handler** style — `Cloudflare.Worker(id, props)` pointing at a separate `export default { async fetch(request, env) }` module. That is alchemy's `cloudflare-worker-async` example; its main `cloudflare-worker` example uses the full API, and that is what a provider example should be showing.

## What changed

The Worker is now a class that is both the resource declaration and the runtime entrypoint (`main: import.meta.filename`), with an Effect handler over `HttpServerRequest` / `HttpServerResponse` — matching `.context/alchemy-effect/examples/cloudflare-worker`.

Three things fall out of that, and they are the reason it is worth the rewrite:

- **`env` accepts an `Effect<Redacted<string>>`**, so `IngestKeys.privateKey` is bound declaratively. The Worker deploy is ordered behind the ingest-key read by the resource graph rather than by statement order, and the value stays `Redacted` in plan output.
- **The impl runs once per isolate**, so the config reads and the HTTP client live there and the per-request handler stays cheap. This is the shape alchemy beta.62 reorganised around ("build layers once per isolate, scope resources per event").
- **`Cloudflare.WorkerExecutionContext.waitUntil` takes an Effect**, so the span export keeps the caller's context and tracing instead of being an orphan promise.

Two API details worth noting for anyone else writing one of these against `alchemy@2.0.0-beta.64`: the bindings prop is `env`, not `bindings` (the `bindings` name is from the older `alchemy-effect` package), and it is `Config.string` / `Config.redacted`, not the capitalised forms.

## Verification

`bun run --cwd examples/alchemy-maple typecheck` and the full `bun typecheck` (33/33) pass. The example gained a tsconfig and a typecheck script in #264, so CI compiles it — which is what caught both mistakes above.

Not deployed: this needs a real `MAPLE_API_KEY` plus a Cloudflare account, so the code is compile-verified only.

---
<a href="https://app.blacksmith.sh/MapleTechLabs/codesmith/maple/pr/265"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787675663&installation_model_id=427786&pr_number=265&repository=MapleTechLabs%2Fmaple&return_to=https%3A%2F%2Fgithub.com%2FMapleTechLabs%2Fmaple%2Fpull%2F265&signature=4fd7bcbdceedb756a5dec12e30f65d7e8ee4c73cf1a5e3b54b7ae3b8994332da"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---

<a href="https://app.devin.ai/review/mapletechlabs/maple/pull/265" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>


Makisuo:
Pushed a second commit: the Worker now uses **`@maple-dev/effect-sdk/cloudflare`** instead of hand-rolling its OTLP/JSON payload (`resourceSpans` literals, hex id generation, nanosecond-string formatting — ~50 lines, all gone).

`MapleCloudflareSDK.make()` at module scope, `telemetry.layer` on the handler, `telemetry.flush(env)` inside `ctx.waitUntil`.

The swap is what makes the ingest-key binding pull its weight: a `Redacted` env value ships as a Cloudflare `secret_text` binding, so at runtime `env.MAPLE_INGEST_KEY` is the plain string the SDK already looks for. The resource output feeds the SDK with nothing in between — and without it the SDK silently runs in no-op mode, which is exactly the failure mode the example should be demonstrating the fix for. `OTEL_SERVICE_NAME` and `MAPLE_ENDPOINT` dropped out of `env`: the service name is passed in config, and the endpoint already defaults to the public ingest.

Two ordering details the code now encodes, both easy to get wrong:

- **The span closes before the flush is scheduled.** `waitUntil` starts the promise immediately, so a span still open at that point misses the drain. (`apps/electric-sync` yields a macrotask before flushing for this reason — it uses `HttpMiddleware.tracer`, which ends the root span deferred. Not needed here.)
- **`telemetry.layer` goes on the `fetch` effect, not the impl.** `WorkerShape` only admits Worker services in a handler's requirements, and the bridge deliberately strips its memo map so a handler-level `Effect.provide` builds per event (`WorkerBridge.ts:337`). Cheap here: the span/log buffers live in `telemetry`, not in the layer.

Also rewrote the README around that seam, and commented `SERVICE_NAME` as the single source of truth it is — the SDK reports it as `service.name` and the alert rules filter on it, so drift means the alerts watch a service that never reports.

Verified: `bun typecheck` 33/33, `bun run --cwd examples/alchemy-maple typecheck` clean, oxlint clean. Compile-verified only — deploying needs a real `MAPLE_API_KEY` and a Cloudflare account.

PR #283, opened 2026-07-30
feat(onboarding): add Bento client and close the MAPLE_ENVIRONMENT gate

Makisuo:
Step 1 of moving the onboarding drip sequence into [Bento](https://bentonow.com), so the welcome → connect-nudge → stalled → activation sequence becomes editable without a deploy.

**Ships dark.** No worker builds the layer, so `BentoService` has no call sites and nothing can send. Alert notifications and the weekly digest stay on the Cloudflare Email Service binding — they're transactional, and the digest renders per-org arrays that Liquid merge tags can't express.

## Why

`OnboardingEmailService` is a hand-rolled drip engine: a daily cron paginates every Clerk org and its memberships to discover recipients, then walks a four-branch ladder against day counters and four `*_sent_at` columns. Copy changes need a deploy, and there are no tests.

## The `MAPLE_ENVIRONMENT` fix

Not strictly part of the Bento work — found while designing containment, and worth reviewing on its own.

`alchemy.run.ts` passed `MAPLE_ENVIRONMENT` through `optionalPlain`, which is `process.env[key]?.trim() || fallback` — so **`process.env` wins over the stage-derived value**. Both non-prod email gates derive from that single variable:

- the alerting worker's `scheduled()` early return
- `EmailService.emailAllowed`

A stray `MAPLE_ENVIRONMENT=production` in a `pr-N` deploy environment therefore opened both at once, leaving the prd-only `EMAIL` binding as the only thing preventing a preview stage from emailing real users. Now derived from the stage and non-overridable, in both email-capable workers.

## BentoService

The two operations are split along their blast radius, and that split is the core safety property:

| Method | Endpoint | Can it email? |
|---|---|---|
| `upsertSubscribers` | `POST /batch/subscribers` | No — documented as not triggering Flows |
| `trackEvent` | `POST /batch/events` | **Yes** — this is what Flows trigger on |

A subscriber backfill necessarily touches every org, so if Flows triggered on "subscriber created" a backfill would mass-mail the user base. Flows must trigger on the event.

Containment, most trustworthy first:

1. **Credential scoping** — prd-only secrets in alchemy, so a non-prod worker never receives them. This is the layer that survives a bad env var, replacing what the prd-only `EMAIL` binding did for Cloudflare.
2. **The `enabled` gate**, re-checked inside every mutating method so a caller holding a handle can't skip it. `MAPLE_BENTO_ALLOW_NONPROD` is deliberately *separate* from `MAPLE_ALERTING_ALLOW_NONPROD` — running the crons on a stage to debug them is routine and must not imply permission to write real contacts.
3. Flow entry conditions in Bento's UI.

Follows `PlanetScaleService`: Effect HttpClient (not `fetch`), `Schema.TaggedErrorClass`, lenient `Schema.fromJsonString` decoding, 15s timeout, `Layer.provide(FetchHttpClient.layer)`. Carries over the PII rule from `EmailService` — recipient addresses are never annotated onto spans.

## Reviewer notes

Three auth details are easy to get wrong and are pinned by tests: HTTP **Basic** with `publishable:secret`, `site_uuid` as a **query parameter**, and a **`User-Agent`** header without which Cloudflare blocks the request outright.

`BENTO_PUBLISHABLE_KEY` is `optionalRedacted`, not `optionalString` — it's the Basic-auth username half and still a credential despite the name.

Secrets go in the Infisical **`prod`** environment only; adding them to `staging`/`dev` has no effect and weakens the containment. Documented in the README.

## Verification

- 8 new tests pass, scoped run: `bun run vitest run src/services/BentoService.test.ts`
- `tsc --noEmit` clean on `apps/api` and `apps/alerting`
- Credentials confirmed against the live Bento site via `infisical run --env=prod --silent -- bun run bento:check` (read-only `GET /fetch/tags` → 200)

## Not in this PR

The schema (`bento_cohort`, the event-claim table), the slimmed tick, and the cutover. Current plan is to **not** backfill existing orgs — the sequence only applies to new signups, and skipping the backfill removes the mass-send risk rather than defending against it. A marketing import is a separate, deliberate piece of work.

---
<a href="https://app.blacksmith.sh/MapleTechLabs/codesmith/maple/pr/283"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788024203&installation_model_id=427786&pr_number=283&repository=MapleTechLabs%2Fmaple&return_to=https%3A%2F%2Fgithub.com%2FMapleTechLabs%2Fmaple%2Fpull%2F283&signature=4afba98243e128af2fcec2a4d2a89f90c07582a7b0d4002ceba7da7dd2fc2dbd"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


Makisuo:
Superseded by #284, which keeps only the `MAPLE_ENVIRONMENT` hardening.

The Bento integration is dropped. Onboarding will instead move to a Cloudflare Workflow with durable `step.sleep` — the Email binding, React Email templates, Postgres state and Workflows infrastructure are already in place, so it's roughly a day's work and removes the daily-tick timing skew rather than inheriting it. A marketing/broadcast tool is a separate follow-up, chosen on its own merits.

The env-var override bug was real and independent of any vendor, so it ships on its own in #284.

PR #286, opened 2026-07-30
perf(ci): cache everything, unserialize typecheck, fan out the archive probes

Makisuo:
## Why

CI was 7–22 min per PR. The turbo step reported this on **every single run**:

```
Tasks:    59 successful, 59 total
Cached:    0 cached, 59 total
Time:    6m1.346s
```

There was no `actions/cache` anywhere in `.github/workflows/` except `Swatinem/rust-cache`. So `.turbo/cache` died with the runner, `bun install` ran cold in ~13 job executions per push, and Playwright re-downloaded three browser engines every run.

Measured baseline (`gh run view --json jobs`):

| Job | Time | Runner |
| --- | --- | --- |
| Local archive native | **7m09s** (worst 21m30s) | `ubuntu-latest` |
| TypeScript | **6m44s** (worst 10m09s) | blacksmith-4vcpu |
| Service Map Perf | 4m13s | blacksmith-4vcpu |

## What changed

**Caching** — `.turbo/cache`, the bun module store (new `.github/actions/bun-install` composite), and Playwright browsers. Blacksmith routes `actions/cache` to its own colocated backend, so no vendor-specific action is needed. `install-deps` stays unconditional: apt paths aren't covered by `~/.cache/ms-playwright`, so the combined `--with-deps` form would skip them on a cache hit and break webkit.

**Turbo graph** — dropped `^typecheck` from `typecheck.dependsOn`. No tsconfig here uses `composite`/`references` and every typecheck is `tsc --noEmit`, which emits nothing downstream — so the edge was ordering-only and serialized a topological wave across ~26 packages. `^build` is the real edge and stays. TypeScript job moved to 8 vCPU with `--concurrency=10`; 15-way concurrency on 4 cores was thrash plus OOM risk once the DAG flattens.

**Archive probes** — 7 sequential probes on `ubuntu-latest` → a 7-way matrix on Blacksmith, timeout 90 → 20 min. Each leg rebuilds the bundle rather than sharing one via artifact: `dist/` is ~403 MB and `build-local-binary.sh` only takes ~14s.

**Path filtering** — `dorny/paths-filter` gates the CLI probes, perf, ClickHouse E2E and slack-agent. Knip stays unfiltered (whole-repo dead-code check). A `CI passed` aggregator collapses the fan-in, since a skipped job never reports a status.

## Correctness fixes found along the way

- `deploy-prd`/`deploy-stg` fired on `push: main` with their own concurrency groups, racing `ci.yml` — **a red CI still shipped to production**. Both now gate on `workflow_run` of CI with `conclusion == 'success'`, pinned to `head_sha`.
- `ci.yml` never typechecked `tsconfig.alchemy.json` (every `apps/*/alchemy.run.ts`) — only the root `typecheck` script chains it, and CI calls turbo directly.
- The Rust base-branch build ran in a `$RUNNER_TEMP` worktree, outside `rust-cache`'s `workspaces: apps/ingest` scope, so it cold-compiled every dependency on every PR.
- 14 of 17 workflows had no `timeout-minutes`, including all three deploys.

## Verifying

The **second** run on this branch is the one that matters — the first only populates the cache. Watch for `Cached: 0 cached` flipping to a high hit count, and confirm `Web bundle budget` still passes against a restored `apps/web/dist`.

---
<a href="https://app.blacksmith.sh/MapleTechLabs/codesmith/maple/pr/286"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788044573&installation_model_id=427786&pr_number=286&repository=MapleTechLabs%2Fmaple&return_to=https%3A%2F%2Fgithub.com%2FMapleTechLabs%2Fmaple%2Fpull%2F286&signature=9d4a35ace48ab34270c5d5263f73b0184390eb0fccbbe26c24c701461ea9b827"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -1,0 +1,7 @@
+# Blacksmith runner labels. actionlint only knows the GitHub-hosted set, so without
+# this every `runs-on:` in the repo reports as an unknown label and drowns out real
+# findings. Add new sizes here when a job starts using one.
+self-hosted-runner:
+    labels:
+        - blacksmith-4vcpu-ubuntu-2404
+        - blacksmith-8vcpu-ubuntu-2404
--- a/.github/actions/bun-install/action.yml
+++ b/.github/actions/bun-install/action.yml
@@ -1,0 +1,28 @@
+name: bun install
+description: >-
+    Restore the bun module store and run a frozen-lockfile install. Every CI job
+    paid a cold 29-48s install before this existed; the store cache brings that
+    down to ~10s. Blacksmith transparently routes actions/cache to its own
+    colocated backend, so no vendor-specific action is needed.
+
+inputs:
+    working-directory:
+        description: Directory holding the bun.lock to install from.
+        required: false
+        default: "."
+
+runs:
+    using: composite
+    steps:
+        - name: Restore bun store
+          uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+          with:
+              # The global module store, not node_modules — bun still relinks per
+              # workspace, which is fast; the expensive part is the network fetch.
+              path: ~/.bun/install/cache
+              key: bun-${{ runner.os }}-${{ hashFiles(format('{0}/bun.lock', inputs.working-directory)) }}
+              restore-keys: bun-${{ runner.os }}-
+
+        - run: bun install --frozen-lockfile
+          shell: bash
+          working-directory: ${{ inputs.working-directory }}
--- a/.github/actions/bun-install/action.yml
+++ b/.github/actions/bun-install/action.yml
@@ -1,9 +1,15 @@
 name: bun install
 description: >-
-    Restore the bun module store and run a frozen-lockfile install. Every CI job
-    paid a cold 29-48s install before this existed; the store cache brings that
-    down to ~10s. Blacksmith transparently routes actions/cache to its own
-    colocated backend, so no vendor-specific action is needed.
+    Frozen-lockfile install, shared by every CI job so there is one place to
+    change install behaviour.
+
+    Deliberately NOT cached. ~/.bun/install/cache for this repo is ~739MB, and
+    caching it measured consistently SLOWER than a cold install: knip's install
+    went 34s -> 57s and local-checkpoint-native's 44s -> 113s, because restoring
+    and re-saving that much data costs more than bun spends fetching. Do not
+    re-add actions/cache here without measuring the step both ways. If this
+    becomes a bottleneck the thing to try is a Blacksmith sticky disk mounted at
+    node_modules, which avoids the tarball round-trip entirely.
 
 inputs:
     working-directory:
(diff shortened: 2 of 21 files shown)


Makisuo:
## Measured results

Verified across 4 runs on this branch. Numbers are the full suite — this PR touches `turbo.json` and `.github/**`, which are in the shared `base` filter anchor, so **every** job ran. A PR scoped to one area will skip more.

| | before | after (warm) |
| --- | --- | --- |
| **Wall clock** | 7m09s – 22m | **4m48s** |
| TypeScript job | 6m44s (worst 10m09s) | **1m17s** |
| turbo step | 6m01s, `0 cached / 59` | **1.8s**, `59 cached / 59` — `>>> FULL TURBO` |
| Local archive native | 7m09s (worst 21m30s) | **2m21s** (slowest of 7 legs) |

Cold-cache runs land at ~6m, so even a first-run-on-a-branch beats the old typical.

The critical path is now `Service Map Perf` at 4m24s, of which 155s is the perf gate itself.

### Two things did not go as planned, and are corrected here

**The bun store cache was a regression and has been removed.** `~/.bun/install/cache` is ~739MB for this repo; restoring and re-saving it cost more than bun spends fetching:

```
knip install              34s cold ->  57s cached
local-checkpoint-native   44s cold -> 113s cached
```

The composite action stays as the single place install behaviour lives, with the measurement recorded in it so this doesn't get re-added on intuition.

**The Playwright cache is a wash, not a win.** Browser install now takes 0s, but the 12s cache restore plus the irreducible 24s apt `install-deps` adds up to about the 38s download it replaced. Kept anyway — it removes a network dependency from the critical path — but it is not where the time went.

### Cache portability confirmed

`Web bundle budget` passes against an `apps/web/dist` restored entirely from cache, with chunk hashes byte-identical to the pre-change baseline (`assets/index-DBM1wOsy.js`). No absolute paths leaked into the artifacts.

### Flake soak

The archive matrix is the riskiest change — crash/GC probes simulate process kills against a `bun --compile` binary talking to chDB over `bun:ffi`, and they moved off `ubuntu-latest` onto Blacksmith. **4 consecutive green runs, 28/28 legs.** The plan called for 5; one short.

### One real bug this surfaced

The new `tsconfig.alchemy.json` check failed on its first run — `CreateAlertingWorkerOptions.chatFlue` was declared `{ workerName: string }` while `createChatFlueWorker` returns a worker whose `workerName` is `Output<string, never>`. The config only typechecked because it was never typechecked. Fixed in 613ed0e.

PR #300, opened 2026-08-01
feat(cli): add versioned local-store migrations

### LINKED ISSUE #297: Local-mode schema changes leave no lossless upgrade path for existing stores
Upgrading a populated local-mode store to a Maple build with a newer bundled schema can cause an outage. Startup rejects the existing store as stale, while the documented recovery requires wiping live telemetry.

The existing store was created by a locally built deployment whose schema came from PR #194 ancestor `8330749a`. The additional deployment patch produced commit `9fb54780`/tree `3f5d7c18`, but did not modify the bundled schema. PR #194 later advanced and merged independently. The relevant destination schema is upstream main at `68be76c0`.

Reproduction:
1. Start an older Maple build with a persistent local data directory.
2. Ingest telemetry.
3. Stop Maple cleanly.
4. Install current `main`, retaining the data directory.
5. Start Maple against that directory.

Startup fails with:
```text
the local store … was built from an older schema. Maple preserved it and its checkpoints; explicitly rebuild live data with `maple start --reset` or `maple reset --yes`.
```

Rolling back to the previous binary restores service and confirms the store remains intact. The chDB version was unchanged, and the failure was unrelated to dirty-store recovery.

`--reset` preserves checkpoints, but checkpoint manifests are also schema-fingerprint-gated, so they do not provide a lossless upgrade path.

Expected: versioned migrations, a side-by-side rebuild/import/swap command, or at minimum a pre-upgrade compatibility check and explicit upgrade warning.

We retained the original store and plan to test a controlled, lossless manual migration. We will post the validated procedure and verification results here afterward.

robbiemu:
## Summary

This draft adds a versioned, fail-closed local-store migration framework for the local-mode schema upgrade gap described in #297.

### What changed

- Adds v2 store markers with stable store identity, creation provenance, full schema digest, activation state, and migration metadata.
- Adds an immutable v1 schema/manifest snapshot so historical migration edges do not silently retarget when the current schema advances.
- Refactors execution into statically registered typed migration modules with `preflight`, `prepareTarget`, `apply`, `verify`, and `recover` handlers. The coordinator owns locking, journal durability, chain progression, staging, promotion, activation, and completed-journal archival.
- Journals the complete migration chain, current step, frozen identities, module version, typed module state, and resumable progress. Each module now supplies runtime decoders, so malformed persisted state fails closed.
- Implements the legacy v0 → v1 side-by-side rebuild in its own module: explicit raw-table columns, retention/cutoff bounds, bounded resumable batches, source/target inventories, physical-schema verification, and rollback-only source retention. Pending batches now validate a SHA-256 signature over every canonical row before recovery accepts them.
- Writes the migration open sentinel before native chDB connect/bootstrap and clears it only after clean close. Dirty populated sources stop fail-closed; staged targets are never silently reopened.
- Adds `maple schema abandon --yes`, which validates only the coordinator-owned journal structure and quarantines an unfinished pre-promotion target without requiring its historical module or decoder, preserving the active source, marker, checkpoints, and rollback artifacts. Promotion-started journals remain resume-only.
- Fixes the verified-step crash window: after transition-specific verification, resume only recommits the staged marker and runs no module lifecycle handler, including `recover()` or mutating target preparation. Verified and completed steps must retain persisted state and progress.
- Routes public resume through journal-bound promotion recovery before marker compatibility checks, including the active-marker-still-staging crash window, and validates target/store/path identity.
- Archives completed journals under each migration root so a completed v0 → v1 transaction cannot block a later v1 → v2 transaction.
- Fixes equal-key replay pagination with cumulative duplicate ordinals, bounded offsets while a group remains open, and strict cursors after group exhaustion.
- Adds a genuine two-edge coordinator/resume/promotion test, explicitly covering the shared staged-store topology used by later modules.
- Rejects non-numeric persisted UInt64 cursor values before they reach interpolated ClickHouse SQL.
- Replaces the mutable schema-digest expectation with append-only structural identity history and a CI gate that preserves base identities and requires registered migration reachability.
- Adds a Linux native end-to-end migration/reopen probe using a stopped historical raw-table fixture, with rebuilt service-namespace/database aggregate assertions; it reports `SKIP` when the platform bundle is unavailable.
- Adds `maple schema status`, `maple schema plan`, and `maple schema migrate`, plus updated local-mode documentation.

## Why

Local mode previously rejected a stale store at startup without a lossless upgrade path. Resetting the store discarded telemetry, while opening the store with an incompatible schema/chDB build was unsafe. This provides a bounded recovery path while refusing to fabricate historical derived state or claim old checkpoints are compatible.

## Validation

- Focused migration/recovery suite: 15 passed, 0 failed, 73 expectations.
- Process-kill sentinel test passed; public promotion recovery is exercised through the normal migration entry point and its dry-run preview.
- ClickHouse schema generation, lint, structural-manifest, append-only history, and migration-graph checks passed.
- Full workspace typecheck passed: 33/33 packages.
- Targeted Oxlint, Oxfmt, shellcheck, shell syntax, and `git diff --check` passed.
- Full CLI suite: 405 passed; 3 listener-address tests could not bind `Bun.serve` port 0 in this sandbox (`EADDRINUSE`).

The native probe is skipped in this macOS environment because no platform-matched `libchdb` is available. Linux CI runs the compiled Maple/chDB migration probe and historical raw-table fixture alongside the checkpoint smoke test.

## Review focus

This is ready for maintainer review. Please review the coordinator/module boundary, runtime state decoders, journal and promotion invariants, target-only abandonment semantics, dirty-store sentinel behavior, append-only schema gate, retention/checkpoint guarantees, and whether the registry/executor shape is a sound foundation for later schema-conversion modules.

Fixes: #297

---
<a href="https://app.blacksmith.sh/MapleTechLabs/codesmith/maple/pr/300"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788197069&installation_model_id=427786&pr_number=300&repository=MapleTechLabs%2Fmaple&return_to=https%3A%2F%2Fgithub.com%2FMapleTechLabs%2Fmaple%2Fpull%2F300&signature=097500dc80d18822541a89721549c38f9c2f3cd94487a56320716742842fff36"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


robbiemu:
Wow that was fast 😀

PR #316, opened 2026-08-02
refactor(ui): unify duplicated formatters, hooks, and detail primitives into @maple/ui

Makisuo:
Second dedup sweep after the clipboard unification (#311), from a codebase-wide hunt for UI patterns copy-pasted across `apps/web`, `apps/local-ui`, and `apps/landing`. Everything here converges on existing `@maple/ui` implementations or extracts byte-identical copies — net **−295 lines**, no intentional visual changes beyond drift removal.

## What changed

**Deleted app-side shadows (correctness)**
- `apps/web/src/lib/format.ts` — a 166-line shadow of `@maple/ui/format`; 7 of its 8 exports had zero importers, and its `formatNumber` capped at `M` while the shared one goes to `T`. The one live consumer (Slack card) now uses `@maple/ui/time-format`.
- Five inline relative-time ladders (GitHub/Slack integration cards, template summary, alert diagnosis, local-ui) → the shared floor ladder. The `Math.round` copies rendered "2m" where the shared one renders "1m" for the same instant; that drift is the point — flagging as a deliberate behavior change.
- `apps/web` duplicate `cn`, `lib/keyboard.ts`, `use-reduced-motion` (×2, now `useMediaQuery("(prefers-reduced-motion: reduce)")`), and landing's dead `lib/format.ts`.

**New shared lib/hooks**
- `lib/span-kind.ts` — span-kind labels (×5 copies, two inside `packages/ui` itself) + span-status badge classes (×3). `span-category.ts` re-exports the labels for existing importers; the new leaf is dependency-free so mobile can adopt it later.
- `lib/keyboard.ts` — `isEditableTarget`/`isDialogOpen`; the trace-timeline's inline copy gains the `SELECT` check it was missing.
- `lib/http.ts` gains `HTTP_METHOD_HEX` as the raw-hex source of truth (the same six hexes lived in four places). The Tailwind class map stays literal — JIT can't scan template strings — with a test pinning the two maps together.
- `hooks/use-debounced-callback` — toolbar, range-filter, and the logs sidebar all hand-rolled the timer ref; two of the three leaked the pending timer on unmount. `use-debounced-value` moves from web unchanged.

**New shared components**
- `ErrorSection` — the error alert banner copy-pasted ×4 (both span detail panels + both log detail views); the `prompt` prop renders the standard "Copy as prompt" button on top of the #311 clipboard stack. `formatErrorPrompt` moves from local-ui (its optional-`operation` API won over web's).
- `SearchInput` — the magnifier+clear block, documented as the non-debounced sibling of `ToolbarSearch`.
- `TableSkeleton` — retires five hand-rolled table loading states (metrics, traces, services, errors-by-type, investigations), keeping each one's real headers/widths so nothing reflows when data lands.
- `DetailRail` moves from web (5 importers) with a new `MetaRow` member absorbing the three byte-identical k8s route copies pixel-for-pixel.
- `useContainerSize` adopted where it was being re-implemented (the pie chart inside `packages/ui`, honeycomb, landing volume chart).

## Reviewer notes

- The exploration flagged `semantic-series-colors`' severity map as "missing the `error` key" — that turned out to be **intentional** (`SPAN_STATUS_COLORS` matches `"error"` first); this PR keeps that behavior and just dedups the maps.
- `service-map-particles` and `flow-view` keep their hand-rolled ResizeObservers on purpose: they're imperative canvas resizes with no React state, and converting them would add re-renders to a perf-sensitive path.
- Also restores `.github/workflows/pullfrog.yml`, which a mis-resolved merge on this branch had re-deleted after main restored it.
- Web's log error banner gains the "Copy as prompt" button local-ui already had — the one intentional (small) feature convergence.

## Deferred follow-ups (found, not done here)

Mobile adoption of the new lib leaves (`apps/mobile` has no `@maple/ui` dep; adding that Metro edge needs a live Expo smoke test) · the log detail panel family (~380 near-identical lines across web/local-ui, now a third smaller thanks to `ErrorSection`/`SearchInput`) · the `SpanDetailPanel` shell merge · the 151-site uppercase-eyebrow sweep · the six divergent stat-tile designs.

## Verification

- 168 `packages/ui` vitest tests pass, including new suites for `keyboard`, `use-debounced-callback` (fake timers), `error-prompt` (snapshot), and the hex/class sync test.
- `packages/ui`, `apps/web`, `apps/local-ui` typecheck clean.
- Landing: `astro check` crashes in node on this toolchain (pre-existing); its edits reuse `@maple/ui` import specifiers already live in the same files.
- Not done: browser spot-checks of the k8s detail pages, span error banner, and table loading states — worth an eyeball before merge.

---
<a href="https://app.blacksmith.sh/MapleTechLabs/codesmith/maple/pr/316"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788297962&installation_model_id=427786&pr_number=316&repository=MapleTechLabs%2Fmaple&return_to=https%3A%2F%2Fgithub.com%2FMapleTechLabs%2Fmaple%2Fpull%2F316&signature=b0c6d1eeab34e29a725764931be774bc6857e04cefbdd945a79b0f8b652e1d3a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


Makisuo:
Heads-up on the two failing checks here (investigated from the query-engine hardening session):

- **pullfrog**: `action failed: router balance exhausted` — the review bot is out of API credit. Not a code issue; re-run once the balance is topped up.
- **Local checkpoint native**: the 20-minute timeout is **not caused by this PR**. The native migration probe (landed with #300, which merged without this job ever running green) hangs on the Linux runner and also carries a deterministic assertion bug — its fixture's `Client` span has an empty `ParentSpanId`, so it counts as an entry span and the namespace aggregate reads 2 where the probe expects 1. Fixes are on #315 (`0a8e3ba61`): fixture parentage corrected, every probe phase now announces itself and runs under `timeout`, and the failure artifact collects the migration evidence dir. Once #315 lands, a rebase/merge of main should turn this job green here.

Makisuo:
Supporting evidence for the reconnect-canary hypothesis, from #315's bounded run (the instrumented probe + evidence upload landed there first):

- The probe stalled at `probe: running maple schema migrate` and hit the 600s bound exactly; `maple` emitted zero output in that window.
- The uploaded migration journal (`maple-store-migration.json`) froze in phase `copying` with **only `logs` in `sourceInventory`** — i.e. the hang struck on an early `chdb_connect` cycle, right where the inventory loop reconnects to the source after target work.
- Same binary, same flow completes in seconds on macOS; and `native-checkpoint-smoke.sh` (one long-lived server connection per process) passes on the very same Linux runner. The per-query connect→close→reconnect alternation in `withDb` is the only thing the hanging path does differently.

Artifact with the full journal + store tree: `native-checkpoint-failure-30769990367` on run 30769990367.

Makisuo:
Cross-run differential that narrows the migration hang further, now that the canary result is in:

- **The canary passing rules out bare reconnects** — 20+ alternating source/target cycles, all <160ms on the same runner.
- **The freeze point moves between runs**: #315's artifact journal froze with only `logs` inventoried; this run's artifact (30770914835) froze after `logs`, `traces`, AND `metrics_sum` were inventoried and copied — hang struck while processing the 4th table. So it's not a particular query or a fixed cycle count; it's racy.
- What the real migration's cycles have that the canary's toy stores don't: **background work at close time** — the copied tables have real parts to merge and live MV pipelines from the inserts. A `chdb_close_conn` racing background merge/MV threads (pthread teardown ordering, Linux-specific) fits everything observed: variable trigger point, macOS unaffected, and the server path (one long-lived connection, no repeated closes) green on the same runner.

Suggested next canary iteration: after each target insert, write enough rows to create multiple parts (or `OPTIMIZE TABLE ... FINAL` in flight) before the close — i.e. make the close race real background activity. If that reproduces, the migration-side mitigation is likely `SYSTEM STOP MERGES` (or a settle/flush wait) before each close, or restructuring to fewer connection cycles.

PR #324, opened 2026-08-03
refactor: extract @maple/query-engine-integrations

Makisuo:
> Stacked on #323. Top of the stack.

## Why

`src/ch/queries/` mixed the generic warehouse layer with vendor integrations and single-feature product surfaces. This moves ~2,900 lines out so the dependency runs one way, and nothing under integrations can be imported by a core query.

**Moved** — Cloudflare ×6, PlanetScale ×2, plus `setup-audit`, `billing-usage`, `internal`.

## The cycle, and what stayed

`session-replays` / `session-events` **deliberately stayed in core**. The MCP tool layer shipped from `@maple/query-engine/observability` imports those builders directly (`observability/session-replays.ts:68`, `session-events.ts:94`), so moving them would have made the two packages circular:

```
query-engine/observability → query-engine-integrations → query-engine
```

That constraint is recorded in the new package's barrel rather than left for the next person to rediscover. I verified the rest are clean: **zero** references from `observability/`, `runtime/`, `execution/` or `caching/` to any Cloudflare or PlanetScale builder.

## Shape

Core exposes what integrations build on as explicit subpaths — `@maple/query-engine/ch/{tables,schema,format,query-helpers}` — rather than widening the main barrel to serve a downstream package. No build step in the new package; it follows the same source-exports convention as `@maple/query-engine`.

```
packages/query-engine-integrations/src/
  cloudflare/   (6 modules + tests)
  planetscale/  (2 modules + tests)
  product/      (setup-audit, billing-usage, internal)
  catalog.ts    (+ __sql_baseline__/integrations.sql)
```

## The catalog had to split too

Same acyclicity reason: core's `sql-catalog.ts` cannot import these builders. The 8 vendor fixtures move to `query-engine-integrations/src/catalog.ts` with their own baseline snapshot and their own structural tenant-scope assertion — so coverage moves rather than disappearing, and `EXEMPT_BUILDERS` in core shrinks accordingly.

## Verification

- **The 8 moved SQL shapes are byte-identical** to what core emitted before the move (compared against `HEAD:…/catalog.sql` programmatically).
- Core's baseline diff is **pure removals** — nothing added, nothing modified — so no remaining query changed.
- 972 core tests + 118 integrations tests pass.
- Full monorepo `bun typecheck`: **34/34** (33 before, +1 new package).
- Consumers: 5 files in `apps/api` repointed; `apps/web` and `apps/local-ui` pick up the dependency without source changes.

## Note

Both design passes I ran argued against a separate package (extra build edge, second barrel, no behaviour change) and preferred subdirectories plus a lint rule. This went ahead as a deliberate call; the cycle above was the one genuine surprise, and it's why the extraction is 11 modules rather than 13.


Makisuo:
Superseded by #325.

This branch was rebased out from under itself: `gh stack rebase` rewrote the stack's SHAs, so once #320/#321/#323 merged, git still saw those rewritten commits as unique here — the PR replayed the entire stack (89 files, 8 commits) instead of just the extraction.

#325 is cut fresh from `main` with only the extraction: **43 files, 2 commits, tree byte-identical to this one.** Nothing lost.

Issue #352, opened 2026-08-05
Self-hosting the full stack outside Cloudflare — working Miniflare-based deployment, offering fixes upstream


tlvenn:
We run Maple fully self-hosted (no Cloudflare, no Tinybird): vanilla ClickHouse per `docs/self-hosted-clickhouse.md` with the prebuilt collector, plus the three workers (`api`, `alerting`, `electric-sync`) running under Miniflare in a single container, Hyperdrive bound to a local Postgres 17, Electric fronted by the electric-sync worker, drizzle migrations on boot, chat/AI-triage on the OpenRouter fallback. It's validated end-to-end (auth → Electric shapes → ClickHouse queries) and lives at [tlvenn/maple](https://github.com/tlvenn/maple) ([`deploy/workerd/`](https://github.com/tlvenn/maple/tree/main/deploy/workerd)).

Along the way we found the in-repo Docker path has drifted: `apps/api/Dockerfile` runs `bun run start` which no longer exists, `docker-compose.yml` references `apps/alerting`/`apps/electric-sync` Dockerfiles that aren't in the tree, and `apps/web/Dockerfile` can't build because `.dockerignore` excludes `packages/browser`/`packages/effect-sdk` and `@maple-dev/clickhouse-builder`'s dist is never built (also `@maple-dev/browser` is imported by the web app but not declared, so it only resolves via hoisting).

If there's interest, I'd like to upstream this in pieces:

1. the web Dockerfile/.dockerignore fixes — happy to open that PR immediately;
2. cleaning up or fixing the stale api/compose Docker references — your call whether those should be fixed or removed;
3. the Miniflare runtime as an official self-host deployment path, with docs.

No hard feelings if (3) isn't the direction you want — even just (1)+(2) would help other self-hosters.

collegeimprovements:
Would be great if there is a blessed self-hosting guide. Ideally all powered by docker-compose. That will exponentially increase adoption. We are also blocked on this part right now in our company.

Makisuo:
Hey @tlvenn this is amazing you selfhost it even though it still quite painful to self host right now!
My goal is it to make self hosting Maple as easy as possible best case making it a one click Selfhost experience, would love to see your fixes upstreamed to make it easier for people.

Instead of miniflare I was thinking of using https://github.com/denoland/celld long term for a bit more performance miniflare is lacking in, in my experience.

If you have any issues/painponts or similar with self hosting etc. feel free to create an issue regardless of what it is or DM me directly

[...]

PR #365, opened 2026-08-08
fix(docker): make the web image build and produce a working self-host SPA

tlvenn:
First slice of the fixes offered in #352 — the smallest set that makes `apps/web/Dockerfile` build from a clean checkout and produce a usable self-host image:

- **Declare `@maple-dev/browser` in `apps/web`** — it's imported by `apps/web/src` (session replay + `main.tsx`) but was never in `package.json`, so it only resolved via workspace hoisting. `turbo prune @maple/web --docker` therefore omits `packages/browser` from the pruned tree and the image build fails at `bun run --cwd packages/browser build` (`ENOENT`). Declaring the dependency fixes prune and is honest about the real dependency graph. (`bun.lock` delta is the one matching line.)
- **Build `lib/clickhouse-builder` before the web bundle** — `@maple/query-engine` imports `@maple-dev/clickhouse-builder/sql` and `/expr`, which are `dist/` subpath exports; without building it first, vite fails with `Rolldown failed to resolve import "@maple-dev/clickhouse-builder/sql"`.
- **Add `VITE_ELECTRIC_SYNC_URL` as a build arg** — Electric shapes are the only read path for dashboards/alerts/error-issue lists, and the URL is baked at build time (`apps/web/src/lib/services/common/electric-sync-url.ts`), so a Docker-built SPA needs a way to set it. Defaults unchanged when unset.

Verified by building the image from a clean checkout of `main` and serving it (we run this image in our self-hosted deployment). Happy to follow up with the api Dockerfile / compose cleanup once you've said whether you'd rather fix or remove those (#352).

---
<a href="https://app.blacksmith.sh/MapleTechLabs/codesmith/maple/pr/365"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788749555&installation_model_id=427786&pr_number=365&repository=MapleTechLabs%2Fmaple&return_to=https%3A%2F%2Fgithub.com%2FMapleTechLabs%2Fmaple%2Fpull%2F365&signature=4bd6bce15f0ae183a64418b81e9bed122136a77ff6b245e87f1870f8a7ec6480"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


Makisuo:
Thank you!!

PR #416, opened 2026-08-11
perf(api): drop the dial budget, retry and probe from the Postgres path

Makisuo:
**350 insertions, 503 deletions.** This removes three mechanisms rather than adding a fourth.

## Why

Production kept emitting `CONNECT_TIMEOUT` after #408, tagged as a warehouse error on the `capabilities` pipe. It isn't a warehouse failure — `"capabilities"` is just the label `resolveRoute` passes (`executor.ts:284`); the real failure is a Postgres read in `OrgClickHouseSettingsService.selectCachedRow`, re-tagged at `WarehouseQueryService.ts:350-360`. The request *was* inside a connection scope and the scope worked. Its one connection failed.

#408 was never going to stop that: it reduced how many connections a healthy request opens, not what happens when one fails.

## The chain being removed

| Commit | Added | Because |
|---|---|---|
| `a61f4d2` | 2s dial cap | dials stalled to Hyperdrive's 15s; p99 was 15.4s |
| `aa67037` | retry at 8s | the cap alone took 5xx from 0.06% → **5.01%** |
| `d967e4e` | one connection per request | requests opened 4–6 |
| *(next would have been)* | failure cooldown | the retry re-arms per sibling, so failures fan out |

Each layer fixes the previous layer's side effect. Cloudflare's Hyperdrive example is one lazily-created client per request; #408 already converged on that, and everything stacked above it was ours.

## What dropping the probe unlocks

The `select 1` probe existed to synthesize a connect phase so `db.connect_ms`/`db.query_ms` could be split — at the cost of a round trip on every DB-touching request. Without it, **postgres.js connects on the first statement, so creating a client is synchronous and cannot fail.** That removes the async dial, the single-flight promise, and any need for a failure latch in one move:

```ts
const acquire = () => (socket ??= openSocket())
```

A connection problem now arrives as the statement's own error, classified by `postgres-errors.ts` — `error.type` states outright what the duration split was being used to infer.

`executeOnFreshPgClient` and `makeTracedPgConnection` shed the same machinery.

## Deliberately kept

Neither belongs to the compensation chain:

- **`max: 1`** — capped by Cloudflare's six-simultaneous-outbound-connection budget (measured with numbers in `lib/cache/src/edge-cache.ts:122-150`), not by anything to do with dialing. Not reverted to the example's `max: 5`.
- **Per-call `wrapMaplePgClient`** — drizzle fixes its logger at construction, so a shared wrapper cross-attributes SQL between concurrent executes.

## Also here, independent of the connection mechanism

- A short **negative cache** on the org-config read, so an unreachable origin isn't re-discovered by every branch of the documented 22-branch in-request fan-out.
- **`forkRequestScoped`** replacing `Effect.forkDetach` at `ApiKeysService.ts:434` and `ScrapeTargetsService.ts:674`. A detached fiber inherits the scope reference but not the request lifetime, so it could outlive the connection it writes through — a latent bug from #408.

## Accepted trade

With no cap and no retry, a bad connection can stall toward Hyperdrive's timeout instead of failing at 10s. `a61f4d2` measured that under dial-per-execute, so the exposure is now **one** connection per request rather than four to six. The explicit intent is to simplify, then measure — nothing has been measured since #408 deployed. `aa67037`'s 5.01% figure was *cap-without-retry*, a different configuration, and shouldn't be read as a prediction here.

## Verification

- `bun typecheck` 37/37; `lint:effect` clean.
- `@maple/api` 1677 passed, `@maple/db` 5 passed.
- Real-Postgres integration suite 5/5 (runs in the `Postgres connection scope E2E` CI job).
- **Mutation-checked**: sharing one drizzle wrapper fails the cross-attribution test with `beta_marker` appearing on alpha's span; raising `max` above 1 fails the `pg_stat_activity` backend count.
- One integration assertion genuinely changed meaning and was fixed rather than forced: sampling backends *inside* the first execute now legitimately reads 0, because the client is lazy. It samples after each execute instead.

Pre-existing failures unrelated to this change: `@maple/cli` (needs `/usr/bin/time` + IPv6), `@maple/domain` (`project-manifest.test.ts` 5s timeout), `@maple/alerting` (`vitest: command not found` in my container only).

## Follow-up worth settling

`61314afdf90bb5d91780b0244ff87901` — the Hyperdrive host in the production error — appears nowhere in the repo, yet `stage.ts:191-200` hardcodes `ad4c4878…` for prd api and `f4731672…` for prd alerting. Worth querying `Database.execute` spans by `server.address` / `service.name` / `deployment.environment.name` to confirm which worker and stage that host belongs to. If prd api is bound to an unlisted config, the api/alerting pool split from `0b27213` may not be in effect.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MbizSpW9vjfVRrBKZvwHPS)_

---
<a href="https://app.blacksmith.sh/MapleTechLabs/codesmith/maple/pr/416"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789042590&installation_model_id=427786&pr_number=416&repository=MapleTechLabs%2Fmaple&return_to=https%3A%2F%2Fgithub.com%2FMapleTechLabs%2Fmaple%2Fpull%2F416&signature=2d8f2ba4681b58aecab2d6216d3ef36742f917af59a3e4c2a40ab511dfd8762c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/apps/api/src/platform/DatabaseLive.test.ts
+++ b/apps/api/src/platform/DatabaseLive.test.ts
@@ -187,58 +187,44 @@ describe("Database execute span instrumentation", () => {
  * a slow query — the ambiguity that made the production p95 investigation
  * guesswork. These cover the split that resolves it.
  */
-describe("Database execute connect/query split", () => {
-	it.effect("reports both phases when the body marks the handshake", () =>
-		Effect.gen(function* () {
-			const { spans, tracer } = makeRecordingTracer()
-			const database = yield* Database
-
-			yield* database
-				.execute((db) => db.execute(sql`select 1 as probe`))
-				.pipe(Effect.withTracer(tracer))
-
-			const [span] = dbSpans(spans)
-			assert.isDefined(span)
-			assert.strictEqual(span.attributes.get("db.connect.completed"), true)
-			assert.isNumber(span.attributes.get("db.connect_ms"))
-			assert.isNumber(span.attributes.get("db.query_ms"))
-		}).pipe(Effect.provide(createTestDb(trackedDbs).layer)),
-	)
-
-	it.effect("attributes time before the handshake to connect, not to the query", () =>
-		Effect.gen(function* () {
-			const { spans, tracer } = makeRecordingTracer()
-
-			yield* executeWithSpan(async (hooks) => {
-				await new Promise((resolve) => setTimeout(resolve, 30))
-				hooks.markConnected()
-				hooks.collect("select 1")
-				return "ok"
-			}).pipe(Effect.withTracer(tracer))
-
-			const [span] = dbSpans(spans)
-			assert.isDefined(span)
-			// Margin for timer imprecision; the point is that the pre-mark wait
-			// landed in connect instead of being folded into query time.
-			assert.isAtLeast(span.attributes.get("db.connect_ms") as number, 25)
-		}),
-	)
-
-	it.effect("flags a dial that never completed and omits query time", () =>
+describe("Database execute failure classification", () => {
+	// There is no connect/query split any more. postgres.js connects on the first
+	// statement, so the split could only ever be synthesized by a `select 1` probe
+	// that cost a round trip on every request. What it was used to infer — is this
+	// a connection problem or a query problem — `error.type` states outright.
+	it.effect("classifies a connection failure by class rather than by duration", () =>
 		Effect.gen(function* () {
 			const { spans, tracer } = makeRecordingTracer()
 
 			const exit = yield* executeWithSpan(() =>
-				Promise.reject(new Error("write CONNECT_TIMEOUT")),
+				Promise.reject(
+					Object.assign(new Error("write CONNECT_TIMEOUT"), { code: "CONNECT_TIMEOUT" }),
+				),
 			).pipe(Effect.withTracer(tracer), Effect.exit)
 
 			assert.isTrue(Exit.isFailure(exit))
 			const [span] = dbSpans(spans)
 			assert.isDefined(span)
-			assert.strictEqual(span.attributes.get("db.connect.completed"), false)
-			// Still reports the time burned before giving up.
-			assert.isNumber(span.attributes.get("db.connect_ms"))
-			assert.isUndefined(span.attributes.get("db.query_ms"))
+			assert.strictEqual(span.attributes.get("error.type"), "CONNECT_TIMEOUT")
+			assert.strictEqual(span.attributes.get("db.connect.failed"), true)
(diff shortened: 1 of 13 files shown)


Makisuo:
Superseded by #418 — same commit (`ab0cc73`), unchanged, on a fresh branch.

GitHub never scheduled a check suite for this PR. Both SHAs sat at zero checks while `[code]smith` fired normally on each, so webhooks were working and only the `CI` workflow was never queued. I tried a plain push, an amend + force-push to a new SHA, and a full close/reopen; none produced a run.

The cause is the branch name: `claude/api-database-performance-debug-gcciu7` was deleted when #408 merged and then recreated. #417, pushed to a name that had never existed, triggered all 23 checks immediately — same repo, same workflow, same actor, minutes apart. That's the difference.

Closing this rather than leaving a permanently pending PR. No code review is lost; #418 carries the identical diff and description.

---
_Generated by [Claude Code](https://claude.ai/code)_

Makisuo:
Correcting my comment above: the branch-name explanation was wrong.

The real cause was a **merge conflict with `main`**. `pull_request` workflows run against a merge commit, and GitHub cannot compute one when the PR conflicts — so no run is ever created. That matches what we saw exactly: `total_count: 0` for the branch, rather than a queued or cancelled run.

`a8cbd35e "harden Effect error and telemetry boundaries"` landed on main after this branch's base and edited `apps/api/src/services/org/ApiKeysService.ts`, the one file this change also touched.

#417 ran fine not because its branch name was new, but because it branched off current main and didn't conflict. Recreating this change under a fresh branch name (#418) changed nothing — still zero runs — which is what ruled the branch-name theory out. Once I merged main into #418 and resolved that one conflict, CI was created within seconds (run `31492830624`, 12:46:17).

#418 is merged as `0a5e7b4`. Leaving this closed.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #427, opened 2026-08-11
perf(api): replace autumn-js with an Effect-native Autumn client

JeremyFunk:
## Why

`autumn-js` exists in `apps/api` to do six POSTs against Autumn's REST API — but it drags a Speakeasy-generated SDK whose eagerly-constructed Zod schemas cost ~140MB of heap allocation and real cold-start time on Workers, which is why the import had to be lazy-loaded. This replaces it with a ~300-line Effect-native client and removes the dependency (and Zod with it) from `apps/api` entirely. No lazy-load indirection needed anymore.

## What

- **New [`autumn-http.ts`](apps/api/src/services/billing/autumn-http.ts)** — direct REST calls via `effect/unstable/http`, same house style as the pre-existing hand-rolled `customers.update`. Typed per-route functions: `getOrCreateCustomer`, `aggregateEvents`, `attach`, `previewAttach`, `openCustomerPortal`, `listPlans`, plus `updateCustomerBillingControls` (moved here so all upstream Autumn I/O lives in one module).
- **Wire fidelity**: request bodies are byte-for-byte what the SDK sends today, including its Zod-injected defaults (`expand` always gets `balances.feature` appended, `redirect_mode: "if_required"` on attach/previewAttach, `bin_size: "day"` on aggregateEvents) and the `x-api-version: 2.3.0` header the SDK's env schema defaulted in — all asserted with exact `deepStrictEqual` in tests. Responses go through a deep snake→camel transform matching the SDK's remap semantics exactly: one-level key skip for remapped-model records (`balances`, `flags`, `total`), full-subtree skip for free-form records the SDK never descended into (`metadata`, `properties`, `values`, `grouped_values`).
- **`autumn-client.ts`** shrinks to the cache/decode plumbing (`readCustomerCached`, `ensureOk`, `decodeUpstream`, …), unchanged behaviourally.
- **`autumn-js` removed from `apps/api/package.json`** (kept in `apps/landing` — build-time Astro pricing, doesn't affect worker cold start; `atmn` devDep also kept).
- **Domain schema hardening**: `BillingSpendLimit.enabled` / `BillingUsageAlert.enabled` now carry the SDK's inbound decode defaults (false / true) via `withDecodingDefaultKey`, so Autumn omitting the key can't 502 the page-load path.

## Equivalence audit

A second-pass adversarial audit (static read of `autumn-js@1.2.51` dist — all 1026 remap tables, all 257 inbound schemas reachable from our six response roots — plus a live differential harness running the real `autumnHandler` against a stubbed fetch across 17 scenarios) verified request/response equivalence codepath-by-codepath. Divergences it found (missing `x-api-version`, missing inbound `enabled` defaults, over-deep camelization of `metadata`/`grouped_values`, silent `{}` on empty 2xx) are all fixed in the second commit.

## Intentional behaviour changes

1. **Fail-open removed.** The SDK rewrote a 5xx from `getOrCreateCustomer` into a synthetic 200 with a stub customer — which then failed its own inbound validation, surfacing as a confusing 502 *and* poisoning the edge cache for 5s. Now the 5xx flows through as a non-200 result: `readCustomerCached` refuses to cache it, `ensureOk` 502s with Autumn's own message. Same for the SDK's fake `555 Network Error` on a rejected fetch — now a real `BillingUpstreamError`.
2. **`AUTUMN_API_URL` is now honoured by all routes.** The SDK hardcoded `https://api.useautumn.com` for the RPC routes; the `Env.ts` default is that same URL, so behaviour is identical unless a stage overrides it — worth a glance at the Infisical slugs before merge.
3. **`customerData` on attach dropped explicitly.** `autumnHandler` silently discarded it (`/v1/billing.attach` has no identity fields in API 2.3.0), so the Clerk-identity enrichment documented at the attach handler was never happening. Removing `resolveCustomerData` also drops a wasted `auth.getCustomerData` call from the checkout path; pre-identifying the buyer properly would be a separate `customers.get_or_create` on that path (noted in a comment).
4. **Invalid `range` on getUsage** now returns Autumn's 400 instead of the SDK's Zod-500.
5. **Degenerate-response handling is stricter, not looser**: empty 2xx body → explicit `BillingUpstreamError` (SDK: internal 500→502; new code refuses to cache or fake-succeed); 201/2xx-non-200 and 200-with-wrong-content-type now parse correctly (the SDK turned both into error bodies, one of which it *cached*). Error-body fallback messages are cleaner (`bad thing` instead of `API error occurred: Status 400. Body: {...}`).
6. **Headers**: the SDK's `user-agent: speakeasy-sdk/...` and inert `cookie: ""` are gone. Only relevant if Autumn segments traffic by SDK identity.

## Verification

- `bun run test --filter=@maple/api`: 148 files passed (4 skipped), 1715 tests, 0 failed
- `bun run test --filter=@maple/domain`: 37 files, 483 tests, 0 failed
- `bun typecheck`: 37/37
- knip: no new unused exports; oxlint (both configs) clean
- New unit tests pin the wire format per route: exact bodies, headers incl. `x-api-version`, camelization semantics for both skip classes, non-2xx passthrough, empty-body 2xx, 204, network failure → `BillingUpstreamError`
- Full `/effect-review-v4` multi-agent review ran on the branch (5 concern reviewers + test reviewer + equivalence verifier); remaining findings are non-blocking follow-ups listed below

## Known gaps / follow-ups

- No client-side timeout (parity with the SDK, which also had none) — a ~10s `Effect.timeoutOrElse` would be a sensible follow-up on Workers.
- `BillingUpstreamError` carries only a message: the transport cause chain and upstream HTTP status aren't preserved on the typed error, so callers can't distinguish 401/402/429/503 or DNS-vs-TLS failures without reading logs. Adding `cause`/`upstreamStatus` fields (house pattern: `WarehouseUpstreamError`) is the natural next step.
- A non-2xx Autumn response is returned as a success value, so the `autumn.request` span stays green during an Autumn outage; no structured log is emitted either. Worth a `Effect.logWarning` + span status on that branch.
- `AutumnClient` is a plain factory; promoting it to a `Context.Service` with a `layer` static (house pattern: `PlanetScaleOAuthService`) would move the config check to layer build and open a seam for `SpanNameGenerator`.
- `AuthService.getCustomerData` / `makeGetCustomerData` are now dead in production (only consumer was the deleted `resolveCustomerData`) — sweep or keep for the attach-identity follow-up.
- `apps/web/src/lib/billing/plan-gating.ts` still carries the stale comment/defence for the SDK's fail-open synthetic-200 body, which this PR stops producing.
- The SDK's Zod also *stripped* unknown/odd response fields; our Effect schemas ignore extras but a never-before-seen shape now fails in `decodeUpstream` instead of inside the SDK — same 502, different message. Capturing sandbox fixtures for `customers.get_or_create` / `plans.list` is cheap insurance.
- `apps/landing` still depends on `autumn-js` (build-time only) — separate scope if we want it gone repo-wide.


JeremyFunk:
## Local end-to-end verification ✅

Ran the stack from the worktree (`bun dev` → `worktree-effect-autumn-client.{web,api}.localhost`) signed in with the dev Clerk account, and exercised the billing settings page against the **real Autumn API** with the new client (autumn-js not installed):

- **Plan header**: Starter (Legacy) · Active · Jul 13 – Aug 13 2026 — `customers.get_or_create` + `subscriptions.plan` expand decode
- **Usage by feature**: Logs 3.75 MB / Traces 212.58 MB / Metrics 7.59 MB / Browser Sessions 24 — `events.aggregate`, and `balances`/feature-id keys (`browser_sessions`) surviving the camelize skip-list on live data
- **Billing controls**: per-feature spend-limit rows render ("Paid overage uncapped", "No enforced stop this cycle") — the `enabled`-default decode path
- **Invoices**: 5 invoices with dates, Paid badges, amounts — `expand: ["invoices"]` + `decodeInvoices`
- **Plan catalog**: Startup $39/mo card with allotments + trial CTA — `plans.list`
- **API responses**: `/api/billing/{customer,plans,usage,daily-spend,invoices}` all **200**
- **Outbound spans** (worker trace capture): `POST https://api.useautumn.com/v1/customers.get_or_create` ×2, `/v1/plans.list`, `/v1/events.aggregate` — all **200**, 599–782 ms

Also in this branch (`1a8cdfe`): `bun dev` no longer demands sudo — portless assumed any port < 1024 needs root and shelled out to `sudo` before attempting the bind, which macOS hasn't required since 10.14. Patched via `bun patch` (`patches/portless@0.14.0.patch`, darwin-only threshold change); the proxy now binds 443 as a regular user.

JeremyFunk:
Dropped `1a8cdfe` (the portless sudo patch) per review feedback and force-pushed — the branch is back to the two billing-client commits only. The sudo prompt was a latent portless behavior unmasked by the Aug 3 reboot, not a regression; the long-lived user-owned proxy on 443 is running again, so `bun dev` behaves exactly as before.

PR #430, opened 2026-08-11
feat(domain): vendor the AI vendor classification registry artifact

JeremyFunk:
## What this layer contains

The vendored classification artifact and nothing that reads it:

- `packages/domain/src/ai-registry/registry.json` — the compiled per-span AI/agent telemetry classification registry (21 vendors, 88 matchers, 29 session candidates, a 5-rule unknown tier).
- `UPSTREAM.json` — the trace-capture pin (`cb29d84`) and the artifact sha256, so drift is detectable without a diff.
- `README.md` — provenance, the four compile-time merge decisions, and why the capture corpus is deliberately not vendored.
- `.oxfmtrc.jsonc` — ignores `registry.json`.

## Why it is shaped this way

The artifact is ~280 KB of generated JSON with no logic in it, so it lands alone: a reviewer can diff the data once here and never have to re-read it while reviewing the six behavioural layers stacked on top. It is a build output that happens to be checked in — trace-capture is the source of truth, and patching the artifact in maple to fix a classification bug is the thing the README exists to prevent. The oxfmt ignore is part of that: it is copied verbatim from upstream, and formatting it would turn every future resync into a whole-file false diff.

The README's "TypeScript consumers" section is deliberately absent here — those files do not exist yet. It arrives with them in the fourth PR of this stack.

## Gates run at this level

- `bun run --cwd packages/domain typecheck` — clean.
- `oxlint` and `oxfmt --list-different` over the touched paths — no findings, no formatting differences.

## Dependencies

Based on `main`. Nothing in the stack below it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

---
[...]

PR #431, opened 2026-08-11
feat(api): provision a per-org session-key salt

JeremyFunk:
## What this layer contains

The per-org secret that AI session-key hashing will need:

- `packages/db` — `session_salt_ciphertext` / `session_salt_iv` / `session_salt_tag` on `org_ingest_keys`, migration `0040_org_session_salt.sql` plus its drizzle snapshot/journal.
- `apps/api` — `OrgIngestKeysService.getSessionSalt`, which lazily provisions and decrypts the salt using the same AES-256-GCM envelope (and the same `MAPLE_INGEST_KEY_ENCRYPTION_KEY`) as the private ingest key, with tests.

## Why it is shaped this way

Session keys are stored as `cityHash64(concat(salt, '\0', value))` so the raw key never reaches the warehouse and hashes cannot be correlated across orgs. That makes the salt a prerequisite of the write path rather than part of it, which is why it ships as its own reviewable unit ahead of the classifier.

Two properties are deliberate. The columns are nullable only because existing rows predate them, so a null means "not provisioned yet" and is backfilled on first read — never "this org has no salt". And there is no reroll path: rotating the salt changes every hash, which breaks `SessionsApprox` HLL continuity across the rotation boundary, double-counting sessions on either side of it.

## Gates run at this level

- `bun typecheck` — 37/37 tasks pass.
- `apps/api` vitest, `src/services/org/` — 5 files, 62 tests pass.
- `apps/api` vitest, `src/routes/v2/` — 12 files, 142 tests pass (covers the `v2-test-support` stub gaining `getSessionSalt`).

## Dependencies

Stacked on #430 for ordering only — it shares no code with the registry artifact and is independently reviewable.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

---
<a href="https://app.blacksmith.sh/MapleTechLabs/codesmith/maple/pr/431"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789072262&installation_model_id=427786&pr_number=431&repository=MapleTechLabs%2Fmaple&return_to=https%3A%2F%2Fgithub.com%2FMapleTechLabs%2Fmaple%2Fpull%2F431&signature=3a302ce6eac3d290d42ac67788f34aa2eee3c8b4001a941ac17d8a4eb41ccb28"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/apps/api/src/routes/v2/v2-test-support.ts
+++ b/apps/api/src/routes/v2/v2-test-support.ts
@@ -202,6 +202,7 @@ export const ConfigResourceServiceStubsLayer = Layer.mergeAll(
 	}),
 	Layer.succeed(OrgIngestKeysService, {
 		getOrCreate: die,
+		getSessionSalt: die,
 		rerollPublic: die,
 		rerollPrivate: die,
 		resolveIngestKey: die,
--- a/apps/api/src/services/org/OrgIngestKeysService.test.ts
+++ b/apps/api/src/services/org/OrgIngestKeysService.test.ts
@@ -1,11 +1,12 @@
 import { afterEach, assert, describe, it } from "@effect/vitest"
-import { Cause, ConfigProvider, Effect, Exit, Layer, Option, Schema } from "effect"
+import { Cause, ConfigProvider, Effect, Exit, Layer, Option, Redacted, Schema } from "effect"
 import { IngestKeyEncryptionError, IngestKeyPersistenceError, OrgId, UserId } from "@maple/domain/http"
 import { hashIngestKey } from "@maple/db"
-import { Database, DatabaseError } from "@/platform/DatabaseLive"
+import { encryptAes256Gcm } from "@/platform/Crypto"
+import { type DatabaseClient, Database, DatabaseError } from "@/platform/DatabaseLive"
 import { Env } from "@/platform/Env"
 import { OrgIngestKeysService } from "./OrgIngestKeysService"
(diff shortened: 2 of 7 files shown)


JeremyFunk:
Dropping this one. The salt bought no confidentiality: `AiSessionKeyHash` is a trailing column on `traces`, and the identifier it digests stays in the clear in `SpanAttributes` on the same row, so anything that can read the hash can read the source value beside it. The rest of the stack now hashes the bare value with `cityHash64`, which is all the column ever needed — 8 fixed bytes and a numeric input to `uniqCombined` in `service_ai_vendors_hourly`.

That also removes migration 0040, the lazy backfill, the AES-GCM decrypt on the ingest auth path, and the `ai_salt_missing` counter.

PR #434, opened 2026-08-11
feat(ingest): classify AI spans on the write path and store the columns

JeremyFunk:
## What this layer contains

The schema change and the wiring that makes classification actually happen — the largest layer in the stack.

- **Schema:** five defaulted trailing columns on `traces` (`AiVendor`, `AiSessionKeyState`, `AiSessionKeyHash`, `AiRulesVersion`, `AiRollupHour`) plus the `idx_ai_vendor` and `idx_scope_name` skip indexes, as ClickHouse migration `0015` and local-store migration v4 → v5 (local schema version 5, frozen `local-schema-v5.sql`, history entry, manifest check).
- **Write path:** the classifier from #432 wired into `apps/ingest`'s row writer (`telemetry.rs`), metrics and OTLP spans, and into the local-mode encoder in `apps/cli`.
- **Regenerated artifacts:** `clickhouse-schema.ts`, `tinybird-project-manifest.ts`, `local-schema.sql`, `local-inserts.json` and `clickhouse_insert_mappings.rs` — all produced by `bun run clickhouse:schema` + `bun run tinybird:manifest`, never hand-edited.
- **Verification:** the AI-schema assertions added to `WarehouseQueryService.clickhouse.e2e.test.ts`, and the datasource contract test.

## Why it is shaped this way

Every new column is `DEFAULT`-valued, which makes the `ALTER` metadata-only: no `MATERIALIZE`, no `OPTIMIZE`, no whole-table mutation on `traces`, whose 30-day TTL retires unindexed parts on its own. That also keeps rows written before the classifier shipped readable, with `AiRulesVersion = 0` meaning "never examined" as distinct from "examined and found not to be AI". `AiRollupHour` is receive-time-clamped and written unconditionally for every span, classification on or off, so a skewed client cannot open a partition in 2038.

## ⚠️ Operational flag: BYO-ClickHouse routing

`0015` is `requiredForIngest: true`, so **`clickHouseSchemaVersion` moves from `13` to `15`**. The gateway's INSERT now names all five AI columns, and a BYO cluster without them would reject every direct insert. Orgs stamped below 15 therefore resolve `clickhouse_ready = false` and **route to the managed pipeline until their schema syncs 0015**. That fallback is the designed behaviour rather than a regression, but it is the reason this PR should not merge ahead of the schema rollout.

## Gates run at this level

- `bun typecheck` — 37/37 tasks pass.
[...]

PR #433, opened 2026-08-11
feat(domain): add the typed registry view and its ClickHouse SQL compiler

JeremyFunk:
## What this layer contains

Maple's typed view of the registry artifact, under the `@maple/domain/ai-registry` subpath export:

- `schema.ts` — Effect Schema for the whole document; `registry.json` is decoded at module load and exported deep-frozen.
- `validate.ts` / `validate.test.ts` — the semantic invariants Schema cannot express (globally unique integer priorities, the D4 priority bands, the closed slug set with `unknown:` reserved, `value_prefix` confined to pseudo-keys, justification on sufficient resource matchers).
- `compile-sql.ts` and its unit + ClickHouse e2e tests — lowers the registry to ClickHouse expressions over a `traces` row.
- `registry-fixtures.ts`, `index.ts`, the `./ai-registry` export in `package.json`, `resolveJsonModule` in `tsconfig.json`, and the README's "TypeScript consumers" section.

## Why it is shaped this way

Decoding at module load means a resync that changes the artifact's shape fails the import rather than the first classification, and `validate.test.ts` is the gate a future resync must pass — it reports violations and never repairs them, because a violation means the upstream compiler produced a bad artifact and must be fixed in trace-capture.

The SQL compiler exists for two consumers that both need to evaluate the registry *without* the Rust classifier: the differential tests that prove the two evaluators agree, and the rollup rebuild job, which cannot read the stored `AiVendor` because that is the column it is repairing. It emits no hash SQL, only because nothing in v1 consumes one: `sessionKeyValueExpr` yields the winning candidate's raw value and `cityHash64` over it is a one-liner the caller can write. The subpath export keeps the ~280 KB artifact off the root `@maple/domain` barrel that web and cli import.

## Gates run at this level

- `bun typecheck` — 37/37 tasks pass.
- `packages/domain` vitest — 532 passed, 7 skipped.
- `compile-sql.clickhouse.e2e.test.ts` under `CLICKHOUSE_E2E=1` against a real ClickHouse — 7 passed.

## Dependencies

Requires #430 for `registry.json`. Independent of the Rust layer in #432 — the two evaluators are only tied together in #435.

**Partition note:** `index.ts` here exports everything except the rollup-enablement constants; that module belongs to the rollup and its export line is restored in #436.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

[...]

PR #432, opened 2026-08-11
feat(ingest): add the pure AI span classifier and registry loader

JeremyFunk:
## What this layer contains

Three self-contained Rust modules in `apps/ingest`, plus their tests and benchmark:

- `ai_registry.rs` — loads and validates `registry.json` (via `include_str!`) at first use.
- `ai_classifier.rs` — per-span vendor and session-key resolution over the registry's four-operator algebra, with the resource/scope context hoisted once per batch.
- `cityhash102.rs` — the CityHash64 v1.0.2 variant ClickHouse's `cityHash64` implements.
- `ai_classifier_corpus_test.rs`, a criterion group, and `serde_yaml` as a dev-dependency.

The only edit outside new files is widening `telemetry::any_value_string` to `pub(crate)`.

## Why it is shaped this way

Nothing calls any of this yet. It is pure CPU with no pipeline wiring, no I/O and no schema change, which is exactly what makes a 3,000-line classifier reviewable — its behaviour is fully determined by the registry plus its inputs. The registry is validated at load rather than at classification so a bad resync fails the process start, not the first AI span. The `pub(crate)` widening is the one deliberate coupling: the classifier must canonicalize attribute values through the *same* function the row writer uses, or the Rust and SQL evaluators cannot be proven equivalent later.

The corpus replay is an on-demand local gate driven by `TRACE_CAPTURE_DIR` and skips when the variable is unset, because the ~60 OTLP capture directories are deliberately not vendored.

## Gates run at this level

- `cargo test --locked` — 125 lib tests, 69 bin tests, 0 failures.
- `cargo test --locked` with `TRACE_CAPTURE_DIR` — corpus replay green: **41/41 goldens over 41 captures, 1,514 spans**.
- `cargo clippy --all-targets` — no new warnings; every remaining warning traces to a file this PR does not add (`main.rs`, `telemetry.rs`, `autumn.rs`, `otel.rs`, `usage_metrics.rs`).

## Dependencies

Requires #430 — `ai_registry.rs` `include_str!`s `packages/domain/src/ai-registry/registry.json`.

**Partition note:** the criterion bench's `accept_traces` call stays at `main`'s signature here. Its extra `&AiClassificationSettings::disabled()` argument depends on a `telemetry.rs` type that does not exist until the write-path PR, so that one hunk of the bench file lands there instead. The new `ai_classifier` bench group is complete in this PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

[...]

PR #435, opened 2026-08-11
test(ingest): prove the Rust classifier and the SQL compiler agree

JeremyFunk:
## What this layer contains

The differential proof that the two evaluators of one registry agree — no production behaviour changes.

- `apps/ingest/src/ai_equivalence_fixtures.rs` — drives the real row writer (`encode_traces`) over a branch-covering span set and emits the fixture corpus, plus `telemetry.rs`'s `#[cfg(test)]` hookup for it.
- `packages/domain/src/ai-registry/__fixtures__/equivalence-spans.jsonl` (+ README) — the emitted rows.
- `equivalence-support.ts`, `equivalence.clickhouse.e2e.test.ts`, `hash-alignment.clickhouse.e2e.test.ts`, `corpus-equivalence.clickhouse.e2e.test.ts` — replay those rows through a real ClickHouse and compare the compiled SQL's verdict to the Rust writer's, span for span.

## Why it is shaped this way

The fixtures are generated *from the row writer* rather than hand-written because the thing under test is the writer's canonicalization — first-occurrence-wins duplicate-key resolution, present-but-empty versus absent, `AnyValue → String` coercion. A hand-written fixture would encode our belief about that behaviour instead of the behaviour itself, and would keep agreeing with the SQL after the writer drifted. `fixture_is_reproducible` asserts the emitted bytes are stable, so a drifted classifier fails in Rust before the SQL leg runs at all.

The remaining known divergences are enumerated in `PINNED_DIVERGENCES`, and the suite fails if that set changes in *either* direction — a silently-fixed divergence is as much a signal as a new one. `hash-alignment` separately pins that ClickHouse's `cityHash64(value)` equals the Rust hash, including adversarial non-ASCII shapes and the negative case that multi-argument `cityHash64` is a different function. The fixtures module hooks into `telemetry.rs` rather than its `tests` child because `encode_traces` is private there.

## Gates run at this level

- `cargo test --locked` (+ `TRACE_CAPTURE_DIR`) — 136 lib + 75 bin, 0 failures, including `fixture_is_reproducible` and `fixture_covers_every_branch`.
- `packages/domain` vitest under `CLICKHOUSE_E2E=1` — `equivalence` 6 passed, `hash-alignment` 5 passed.
- `bun typecheck` — 37/37 tasks pass.

`corpus-equivalence.clickhouse.e2e.test.ts` skips without a local trace-capture corpus, by the same on-demand design as the Rust corpus replay.

## Dependencies

Requires #434 — the fixtures drive the wired row writer, and the replay inserts into the `traces` schema that PR adds. Requires #433 for the SQL compiler and #432 for the classifier.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

---
<a href="https://app.blacksmith.sh/MapleTechLabs/codesmith/maple/pr/435"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789072277&installation_model_id=427786&pr_number=435&repository=MapleTechLabs%2Fmaple&return_to=https%3A%2F%2Fgithub.com%2FMapleTechLabs%2Fmaple%2Fpull%2F435&signature=7d31f5a5719ebaf5f97dcfaa33d23d07f652423439140c45004196e49e3ebcad"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


JeremyFunk:
Superseded by the v2 stack: #451 → #452 → #453. The registry-artifact + SQL-equivalence design this stack implemented was dropped in the write-side plan's v2 revision; vendor rules are now plain Rust with the vendored classification fixture as the CI gate.

PR #441, opened 2026-08-12
perf(traces): rework the Timeline waterfall on a ref-based viewport

Makisuo:
Reworks the trace **Timeline** tab so pan and zoom are free, and makes it the default tab.

## Why

The viewport lived in a `useReducer`, so every wheel tick, drag frame and tween frame re-rendered the whole subtree. The worst of it was the minimap: one unvirtualized `div` per span, all of them repainting per frame on a 2000-span trace.

The features were already there (search, WASD, colour-by, marquee zoom, clip chevrons). What was missing was an engine that doesn't go through React on every frame.

## The change

The viewport is out of React. It lives in a ref inside `ViewportController` and reaches the DOM two ways:

1. **CSS custom properties** — the rows container carries `--vp0` (window start, trace-relative ms) and `--vpk` (`100 / visibleMs`); each bar carries `--b0`/`--b1` and positions itself with `left: clamp(-50%, calc((var(--b0) - var(--vp0)) * var(--vpk) * 1%), 150%)`. A gesture writes two properties and the style engine moves every bar.
2. **rAF-coalesced pub/sub** for what CSS can't express — the minimap overlay, the ruler, and the clip chevrons / outside-label side.

**Zero React commits during a gesture**, measured with a `MutationObserver`: 30 pan frames → exactly 30 mutations, all `style` on the rows container, none on any row. A 2500× wheel-zoom left the first bar's DOM node identical.

The alternative was to drive zoom by widening an inner scroller and letting native `scrollLeft` pan. I didn't take it: it would have replaced `clampViewport` and its tests, and capped zoom at the browser's ~33M px element limit (~20ms window on a 7-minute trace). The CSS-var approach keeps the ms model and the 0.1ms floor.

Also in here:

- **Ruler** ticks built with `createElement`, budgeted against the measured column (`MIN_TICK_PX = 56`) instead of a fixed count of 6, plus a gridline layer. New `formatDurationAtStep` derives precision from the tick interval — at a 0.83ms window the old formatter printed `600.5ms` sixteen times.
- **Minimap** split into memoized bars + an imperative overlay; edges hit-tested in px (8px) rather than 2%, which at a 3%-wide viewport had made the left handle unreachable.
- **One coordinate space.** `viewportBounds` is now the single definition of "the whole trace" for the minimap, ruler, default view and Fit — there were four, leaving the minimap ~80px out of step with the ruler. Padding is trailing-only so trace-zero sits flush at the left edge, with slack after the last span for the outside duration labels.
- Sticky label column sized from `--sidebar-w`, replacing the spacer divs and the per-row prop.
- ⌥-click folds a whole subtree; expand/collapse one level (`E` / `⇧E`).

## Bug found along the way

The controller's cleanup cancelled its pending rAF but didn't clear the id ref, so after StrictMode's remount the coalescing guard read the stale id as "already scheduled" and dropped every notification for the rest of the session — the ruler rendered nothing while bars positioned fine, because bars go through the synchronous CSS path. There's a `StrictMode`-wrapped regression test; I confirmed it fails without the fix.

## Reviewer notes

- **`@property` is deliberately not used** on `--vp0`/`--vpk`/`--b0`/`--b1`. Unregistered custom properties substitute as tokens so the `calc()` is evaluated in `double`; registering them as `<number>` stores float32, and at a 0.1ms window the rounding puts bars tens of percent off. There's a comment on `writeTimeSurface` saying so.
- **The Waterfall tab is untouched** but is no longer the landing tab. Its `PixelDurationBar` renders duration as 27 fixed blocks, so every sub-millisecond span lights the same single block — that's what prompted this.
- **`timeline-lab`** is a fixture route (registered in `FIXTURE_PATHS`) for iterating without a live API, following the existing `widget-lab` / `node-lab` convention. Happy to drop it if you'd rather not carry it.
- **Not verified against a real trace.** The API wasn't running locally, so everything was exercised against the synthetic fixture; a deep-linked `?spanId=` through `/traces/$traceId` and the detail panel are unexercised.
- Span-event markers and interleaved log rows are out of scope — neither is in the `spanHierarchyQuery` projection.

## Testing

- 272 tests pass in `@maple/ui` (57 new across `viewport-controller`, `timeline-ticks`, `auto-collapse`, and the slimmed reducer). Every existing `clampViewport` case still passes.
- `bun typecheck` clean across all 37 packages; oxlint and knip clean.
- Browser-verified: ⌘-wheel zoom, marquee zoom, shift-drag pan, WASD, F, `/` search, `?` overlay, zoom-to-span, minimap brush/pan/edge-resize/click-to-centre/double-click-fit, ⌥-click subtree collapse, the level buttons, and sidebar resize.

### DIFF
--- a/apps/api/src/http/server-error-span.test.ts
+++ b/apps/api/src/http/server-error-span.test.ts
@@ -119,8 +119,7 @@ const makeHarness = () => {
 	}
 	const serverSpan = (path: string): EndedSpan => {
 		const span = ended.find(
-			(candidate) =>
-				candidate.kind === "server" && candidate.attributes.get("url.path") === path,
+			(candidate) => candidate.kind === "server" && candidate.attributes.get("url.path") === path,
 		)
 		if (span === undefined) throw new Error(`no server span ended for ${path}`)
 		return span
--- a/apps/web/src/lib/error-messages.ts
+++ b/apps/web/src/lib/error-messages.ts
@@ -476,7 +476,7 @@ const normalizeTaggedError = (value: { _tag: string; [key: string]: unknown }): 
 			technicalMessage: rawStringField(value, "causeMessage") ?? technicalMessage,
 		})
 	}
-	if (tag.endsWith("/UnauthorizedError") || /AuthenticationError$/.test(tag)) {
+	if (tag.endsWith("/UnauthorizedError") || tag.endsWith('AuthenticationError')) {
 		return normalized(value, {
 			category: "authentication",
 			title: "Sign in required",
@@ -503,7 +503,7 @@ const normalizeTaggedError = (value: { _tag: string; [key: string]: unknown }): 
 			technicalMessage,
 		})
 	}
-	if (/NotFoundError$/.test(tag)) {
+	if (tag.endsWith('NotFoundError')) {
 		return normalized(value, {
 			category: "not-found",
 			title: "Not found",
--- a/apps/web/src/routeTree.gen.ts
+++ b/apps/web/src/routeTree.gen.ts
@@ -33,6 +33,7 @@ import { Route as ServiceMapBenchRouteImport } from './routes/service-map-bench'
 import { Route as SettingsRouteImport } from './routes/settings'
 import { Route as SignInRouteImport } from './routes/sign-in'
 import { Route as SignUpRouteImport } from './routes/sign-up'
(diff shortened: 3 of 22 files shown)


Makisuo:
Superseded by #442 — the head branch was renamed, which closed this one.

PR #462, opened 2026-08-14
feat(replays): play video-backed sessions behind a replay engine seam

Makisuo:
## Why

The web player hardcoded `new Replayer(...)`, so rrweb was the only thing it could ever play. Mobile sessions record H.264 segments wrapped in rrweb-*shaped* events — the chunk pipeline carries them untouched, but rrweb finds no DOM to rebuild and renders nothing.

## What

Engine construction, the clock, seeking and letterboxing now sit behind a `ReplayEngine` interface (`apps/web/src/components/replays/engine/`), with two implementations:

- **rrweb** — today's `Replayer` moved over unchanged, including its config, the `applyScale` letterboxing math, and the negative-`baselineTime` clock guard (that quirk is rrweb's, not the player's, so it belongs in the engine).
- **video** — decodes each base64 segment to a Blob URL and drives one `<video>`. Since every segment opens on an IDR keyframe, an offset resolves to (segment, offset-within-segment) exactly; a seek landing in a gap between segments snaps forward to the next real footage. Object URLs are held for the current and next segment only — the base64 already sits in the loader's event array, so a decoded Blob is a second copy.

**Nothing below the context changed.** `replay-player.tsx`, `replay-editor-timeline.tsx`, `replay-timeline.ts`, `session-events-panel.tsx`, `replay-range.ts` and `use-replay-chunk-loader.ts` are byte-identical, and the existing player + transport tests pass **as written** — that's the evidence the refactor preserved behaviour, not just a claim.

## Two bugs found along the way

- **Idle derivation would have destroyed video playback.** It treated every event as an instant, but a video session emits ~one event per 30s segment — so each quiet segment read as a 30s idle band, and skip-idle (on by default) collapsed real footage to ~1s. A segment now spans its duration, so only genuine gaps (the recorder stopping while the app is backgrounded) stay collapsible. Regression test included.
- **Touch produced no click markers.** `deriveMeta` only matched `MouseInteractions.Click` (2), never `TouchStart` (7). Also a latent fix for mobile-*web* rrweb sessions, which have never shown markers either.

Separately, `normalizeEvents` ran `JSON.stringify` over both sides of every adjacent pair; with an MP4 inside each event that serialized hundreds of KB per comparison on every range decode. It now rejects on timestamp + type first.

## Format discriminator

`maple.session.replay_format` (`"rrweb" | "video"`) rides the existing `ResourceAttributes` map, so there is **no warehouse migration and no wire change**: the detail query already ships the whole map as JSON and the route loader already prefetches it, so the engine is chosen before any chunk is fetched. Deliberately did not touch `sessionReplaysListQuery` — that would mean four duplicated projections plus a `catalog.sql` regeneration for a list badge nobody asked for.

An absent key means `rrweb`; every session recorded before the marker existed is a browser recording. The browser SDK stamps `"rrweb"` in `buildSessionMetaRow`'s shared base object — placement matters, since `session_replays` is a ReplacingMergeTree that replaces the whole row.

Seeking needed no changes: `is_checkpoint` comes from the `x-maple-is-checkpoint` ingest header, so an SDK that flags every segment gets exact per-segment anchoring from `checkpointAtOrBefore` for free.

## Scope checks

- `apps/mobile` is unaffected — it only calls `/api/query-engine/*` and `/api/dashboards/`, and never reads `session_replays` or any replay route.
- No new column: an earlier `FirstEventMs` attempt was reverted because the deployed cluster rejected it (`warehouse_schema_drift`) and it tripped the local-CLI schema gate. A map key avoids all of that.

## Testing

144 tests across 13 replay files, plus `browser-session` (109) and the effect-sdk session test. Scoped typecheck clean on `@maple/web` and `@maple/browser-session`.

New coverage: segment extraction and offset→segment resolution (gaps, boundaries, past-the-end, negative), the video idle-band trap, touch vs mouse markers, `replayFormat` fallbacks, and the format marker on both active and ended meta rows.

The `<video>` DOM path is intentionally not unit-tested — jsdom implements neither `URL.createObjectURL` nor `HTMLMediaElement.play/pause`, so such a test would only exercise its own stubs. The segment math is pure and covered directly instead.

## Not done

- Repo-wide `bun run test` / `bun typecheck` were skipped (authored late, scoped commands only) — worth a run in CI/morning.
- No live browser verification: it needs a real video-backed session in the warehouse. The `eventsOverride` provider prop injects a synthetic mobile chunk without touching the network.
- Portrait recordings letterbox inside the existing 16:9 surface and still show the browser-chrome header. Engine work only; portrait polish is a separate change.
- **The mobile SDK must hold up its end**: send `maple.session.replay_format: "video"` in its meta row and `x-maple-is-checkpoint: 1` per segment. Unverified from this repo.

---
<a href="https://app.blacksmith.sh/MapleTechLabs/codesmith/maple/pr/462"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789259214&installation_model_id=427786&pr_number=462&repository=MapleTechLabs%2Fmaple&return_to=https%3A%2F%2Fgithub.com%2FMapleTechLabs%2Fmaple%2Fpull%2F462&signature=3f6c05ec5beae5a61c8d514b4830dabf6393781262279e1ebf67efb1744a41d3"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/apps/web/src/components/replays/engine/replay-engine.ts
+++ b/apps/web/src/components/replays/engine/replay-engine.ts
@@ -1,0 +1,88 @@
+// The replay engine seam
+//
+// The player used to construct `new Replayer(...)` inline, which made rrweb the
+// only thing it could ever play. The iOS SDK records H.264 segments and wraps
+// each one in rrweb-*shaped* events so the chunk pipeline carries them
+// untouched — but there is no DOM to rebuild, so rrweb renders nothing.
+//
+// Everything above this interface (the provider's transport state, the trimmed
+// timeline, markers, idle bands, the chunk loader) is format-agnostic and stays
+// exactly as it was. Everything rrweb-specific lives in `rrweb-engine.ts`;
+// everything video-specific in `video-engine.ts`.
+//
+// The contract is deliberately shaped like the rrweb surface the provider
+// already depended on, so the refactor is behaviour-preserving:
+//
+//   new Replayer(events, {root})   -> ReplayEngineFactory.create({mount, events})
+//   getMetaData().totalTime        -> totalTimeMs
+//   getCurrentTime()               -> getCurrentTimeMs()
+//   play(o) / pause(o?)            -> play(o) / pause(o?)
+//   setConfig({speed})             -> setSpeed(speed)
+//   addEvent(e)                    -> addEvent(e)
+//   destroy()                      -> destroy()
+//   on(Finish) / on(Resize)        -> onFinish / onResize callbacks
+//   .iframe + .wrapper.style       -> fit(container)
+
+/** The recorded viewport, used to letterbox the recording inside the surface. */
+export interface Viewport {
+	readonly width: number
+	readonly height: number
+}
+
+export interface ReplayEngineCreateInput {
+	/** The surface's inner div. The engine owns its contents entirely. */
+	readonly mount: HTMLElement
+	/** The seed events. Later events arrive through `addEvent`, forward-only. */
+	readonly events: ReadonlyArray<unknown>
+	/**
+	 * Viewport to fall back to when the engine can't report its own — derived
+	 * from the stream's `meta` events by `deriveMeta`.
+	 */
+	readonly fallbackViewport: Viewport
+	/** Playback reached the end of the recording. */
+	onFinish(): void
+	/** The recorded viewport changed mid-session; the surface must re-fit. */
+	onResize(): void
+}
+
+export interface ReplayEngine {
+	/** Length of the loaded recording, in real ms. */
+	readonly totalTimeMs: number
+	/**
+	 * Playhead as a real-ms offset from session start, matching the clock the
+	 * trimmed `Timeline` and backend span alignment are built against.
+	 *
+	 * Implementations must never return a negative or non-finite value — the
+	 * provider feeds this straight back into `play()`.
+	 */
+	getCurrentTimeMs(): number
+	play(offsetMs: number): void
+	/** With no offset, hold at the current position. */
+	pause(offsetMs?: number): void
+	setSpeed(speed: number): void
+	/**
+	 * Append an event that arrived after construction.
+	 *
+	 * Forward-only by contract: every trailing event postdates the seed. A
+	 * backward seek rebuilds the engine from the nearest checkpoint instead
+	 * (see `requestSeek` in `use-replay-chunk-loader.ts`).
+	 */
+	addEvent(event: unknown): void
+	/** Fit the recording inside `container`, letterboxed and centred. */
+	fit(container: HTMLElement): void
+	destroy(): void
+}
+
+export interface ReplayEngineFactory {
+	create(input: ReplayEngineCreateInput): ReplayEngine
+}
+
+/**
+ * Which engine plays a session's chunks.
+ *
+ * Carried by the `maple.session.replay_format` resource attribute so the player
+ * can pick an engine from session metadata alone, without downloading a chunk
+ * to find out. An absent attribute means `rrweb` — every session recorded
+ * before the marker existed is a browser recording.
+ */
+export type ReplayFormat = "rrweb" | "video"
--- a/apps/web/src/components/replays/engine/rrweb-engine.ts
+++ b/apps/web/src/components/replays/engine/rrweb-engine.ts
@@ -1,0 +1,107 @@
+import { Replayer } from "@rrweb/replay"
+import { ReplayerEvents } from "@rrweb/types"
+import type { ReplayEngine, ReplayEngineCreateInput, ReplayEngineFactory } from "./replay-engine"
+
+// The rrweb engine — browser recordings.
+//
+// This is the behaviour the player has always had, moved behind the engine
+// interface unchanged. Every quirk documented here was a shipped bug once.
+
+class RrwebEngine implements ReplayEngine {
+	private readonly replayer: Replayer
+	private readonly fallbackViewport: ReplayEngineCreateInput["fallbackViewport"]
+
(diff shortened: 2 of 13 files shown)


Makisuo:
## Validated against the first real video-backed session

Checked production before claiming this works. Found session `7B897078-D857-41A2-82BE-6A42E5313413` — 9 chunks, seq 0-8, 51s, iOS. So the SDK **is** already uploading.

**What matched the spec:** event shape per chunk is `type 4` / `type 5 tag=video` / `type 5 tag=breadcrumb`; payload keys are exactly `base64, container, duration, encoding, frameCount, frameRate, frameRateType, height, left, segmentId, size, top, width`; and **all 9 chunks are flagged `is_checkpoint`**, confirming the per-segment seek anchoring works with no changes to range planning.

**What did not:** segments are 5-6s, not 30s — and more importantly the idle fix in the first commit was **wrong**. The SDK stamps `meta`, `video` and `breadcrumb` with the *same* timestamp, breadcrumb last:

```
chunk 0: 4@0      5/video@0      5/breadcrumb@0
chunk 1: 4@5001   5/video@5001   5/breadcrumb@5001
chunk 2: 4@11007  5/video@11007  5/breadcrumb@11007
```

Advancing the watermark by assignment meant the breadcrumb rewound it from `ts + duration` back to `ts`, so the next segment still measured as a 5001ms gap and **every segment became an idle band anyway** — the exact failure the span fix existed to prevent. My synthetic test missed it because it never placed a same-timestamp sibling after the video event.

Fixed by taking the max, with a test built from the real chunk layout. Confirmed it fails without the fix (`expected [{start: 0, end: 5001}, …] to deeply equal []`) and passes with it. 145 tests green.

## Still not verified

The `<video>` DOM path — Blob decode, src swapping at segment boundaries, `loadedmetadata` seek timing, playback rate — has **never executed**. Only the pure segment/idle math is proven.

That session's meta row has `recorded=true` but `replay_format=''`, so it would still route to rrweb today and render nothing. Nothing in production sets the attribute yet (0 of ~5,100 sessions over 30 days). Playing it end-to-end needs the SDK change first, or a manual attribute backfill on that one session.

Makisuo:
## Ran it end-to-end in a real browser — found and fixed a second bug

Built a fixture to the exact envelope verified against the production session (per chunk `meta`/`video`/`breadcrumb` on the **same** timestamp, 402×874 portrait, 1fps, 5-6s segments, real inter-segment drift of 0/5001/11007/17011ms), with genuine H.264 MP4s so the Blob decode and `<video>` path actually execute. Drove it through the real transport UI with a headed browser.

### Bug found: the engine was being destroyed mid-playback

Playback died ~1.3s in — `VideoEngine.destroy` firing, clock resetting to 0. Root cause is **not** video-specific: the mount effect is keyed on the seed `events` array, but that array was being built *inside* the status memo, which depends on the manifest result. Every manifest change minted a new array (`normalizeEvents(...)` / `[...loader.seedEvents]`) and rebuilt the engine.

This is precisely the failure the mount effect's own comment claims to have fixed — the effect was keyed correctly, the seed just wasn't stable. On rrweb it shows up as an iframe flash and the playhead snapping to 0 every few seconds on a **live session**; the video engine loses playback outright, which is how it became visible. Fixed in `38bd233`, with a test that recomputes the memo without changing its outcome: 2 engine instances before, 1 after.

### Verified working after the fix

| Behaviour | Result |
|---|---|
| Continuous playback | clock tracks wall-clock 1:1 (755 → 12010ms over 12s) |
| Segment swaps | 2 swaps at 5.25s / 11.25s, matching the 5001/11007ms boundaries |
| **Clock continuity across a swap** | `clock 5251` at `vT 0.26` → 5001 + 250 ✓ (and 11007 + 251 = 11258 ✓) |
| Seek accuracy | 80% → **40021ms** vs expected 40021; 10% → 5003; 50% → 25013 — each on the right segment, right within-segment offset |
| Backward seek | correct (0.8 → 0.1 → 0.5) |
| Pause | holds position, video paused |
| Speed | `4×` → `playbackRate 4` |
| End of recording | `finished: true`, button → "Replay" |
| Idle bands | **0** across 9 segments (the monotonic fix holding on real-shaped data) |
| Markers | 1, from the TouchStart |
| Letterboxing | 402×874 portrait centred in the 16:9 surface |
| Console | no JS errors |

Screenshot mid-playback: portrait video letterboxed, `0:25 / 0:50` matching `displayCurrentMs: 25013`, marker dot on the scrubber, 4× active.

### One environment note

The in-app preview pane reports `visibility: hidden`, and browsers auto-pause media in a hidden page (`play → playing → timeupdate → pause`). That produced a convincing false negative before I switched to a headed browser — worth knowing for anyone verifying media UI here.

### Still not covered

The harness (bench route + 1.4MB fixture) is **not** committed — too large, and it needs generated MP4s. Everything it proved is now either a unit test or described above. A real mobile session still cannot be played through the normal route until the SDK sets `maple.session.replay_format: "video"`; that session has `recorded=true` but the format key empty, so it routes to rrweb today.

PR #466, opened 2026-08-14
feat(cli): reach a remote local server behind Cloudflare Access

alan-roe:
## Problem

`MAPLE_LOCAL_URL` can already point the CLI at a remote `maple start` instance, but the local-mode HTTP client sends only `content-type` — there is no way to attach an auth header. Any self-hosted instance behind an authenticating reverse proxy (e.g. a Cloudflare Access tunnel, the natural way to expose a single-binary maple on a homelab box) is unreachable from the CLI.

## Changes

- **`packages/query-engine/src/local.ts`**: the local query client accepts optional extra headers (generic `Record<string,string>`; default unchanged, bundled UI unaffected)
- **`MAPLE_LOCAL_HEADERS`**: comma-separated `Key=Value` pairs (same format as `OTEL_EXPORTER_OTLP_HEADERS`), attached to every local-mode request — e.g. a CF Access service token for machines
- **cloudflared identity flow**: when the resolved local URL is non-loopback https and no Access credential is already configured, the CLI best-effort acquires a user JWT via `cloudflared access token -app=<origin>` and sends it as `cf-access-token`. One-time setup is the standard `cloudflared access login <origin>`; missing binary/token degrades silently
- **Typed error UX**: a 401/403 or a redirect to `*.cloudflareaccess.com` / `/cdn-cgi/access/` fails with a `Schema.TaggedError` explaining both options instead of a JSON parse error
- Docs: new section in `local-mode/cli-reference.md`

## Verification

- Targeted unit tests (header parsing, redirect/401 detection, executor wiring): 8 pass; `tsc --noEmit` clean in `apps/cli` and `packages/query-engine`
- Exercised against a real self-hosted v0.0.18 instance behind Cloudflare Access (service-token and identity-token paths)

---
<a href="https://app.blacksmith.sh/MapleTechLabs/codesmith/maple/pr/466"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789303807&installation_model_id=427786&pr_number=466&repository=MapleTechLabs%2Fmaple&return_to=https%3A%2F%2Fgithub.com%2FMapleTechLabs%2Fmaple%2Fpull%2F466&signature=a60ad3c9d405e1614cb461efa48c3b2d07e1fa970f4d925e6d1a1ba7103e1b48"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/apps/cli/src/core/cloudflare-access.test.ts
+++ b/apps/cli/src/core/cloudflare-access.test.ts
@@ -1,0 +1,30 @@
+import { describe, expect, it } from "vitest"
+import { isCloudflareAccessResponse, shouldUseCloudflareAccess } from "./cloudflare-access"
+
+describe("shouldUseCloudflareAccess", () => {
+	it("uses cloudflared only for non-loopback HTTPS without explicit Access credentials", () => {
+		expect(shouldUseCloudflareAccess("https://maple.example.com", {})).toBe(true)
+		expect(shouldUseCloudflareAccess("http://maple.example.com", {})).toBe(false)
+		expect(shouldUseCloudflareAccess("https://127.0.0.1:4318", {})).toBe(false)
+		expect(shouldUseCloudflareAccess("https://localhost:4318", {})).toBe(false)
+		expect(shouldUseCloudflareAccess("https://maple.example.com", { "CF-Access-Client-Id": "id" })).toBe(
+			false,
+		)
+		expect(shouldUseCloudflareAccess("https://maple.example.com", { "CF-ACCESS-TOKEN": "token" })).toBe(
+			false,
+		)
+	})
+})
+
+describe("isCloudflareAccessResponse", () => {
+	it("recognizes denied responses and Access redirects", () => {
+		expect(isCloudflareAccessResponse(401, null)).toBe(true)
+		expect(isCloudflareAccessResponse(403, null)).toBe(true)
+		expect(
+			isCloudflareAccessResponse(302, "https://example.cloudflareaccess.com/cdn-cgi/access/login"),
+		).toBe(true)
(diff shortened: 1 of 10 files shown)


Makisuo:
Hey @alan-roe, this looks really good but I have some small points we should address before merging this in!


- maple sql still broken behind Access — executeRawLocalQuery (operations.ts:327) has no headers, no Access detection. Only the executor path got wired.
- isCloudflareAccessResponse treats any 401/403 as Access. serve.ts:589 "browser origin not allowed" covers /local/query, so a local CORS 403 tells you to run cloudflared access login http://127.0.0.1:4318. Gate it on non-loopback https.
- probeLocal sends a bare /health, so auto-detect never picks a protected origin --local always required.
- no timeout on the cloudflared spawn, hangs the CLI with nothing on screen.
- redirect: "manual" hits apps/local-ui too. Make it a param, default follow.
- unguarded new URL in cloudflareAccessError; headers spread after content-type so they can clobber it; tryPromise catch + orElseSucceed is doubled; const process = shadows the global.

Also I'm curious to what your main use case is of using cloudflare access with it?

